PR #62, opened 2015-11-23
Pins version of spinnaker-dependencies manifest file

cfieber:
Uses the new functionality in the spinnaker-gradle-project for pinning the version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ buildscript {
     maven { url "http://dl.bintray.com/spinnaker/gradle" }
   }
   dependencies {
-    classpath 'com.netflix.spinnaker.gradle:spinnaker-gradle-project:3.1.0'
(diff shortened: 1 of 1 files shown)


ajordens:
Can you make it 0.19.0, it's been fixed in that version.

cfieber:
This pins the dependencies (current behaviour was to pick up latest.release all the time)

Can override at the command-line (`./gradlew build -PdependenciesVersion=latest.release`) which will allow for developing a test suite that we can run after publishing a new version of spinnaker-dependencies

cfieber:
going to look at rolling this block back into the gradle plugin instead of C&P'ing it everywhere

cfieber:
@spinnaker/reviewers FYI - version of the spinnaker-dependencies manifest is now managed this way per project, will apply this change across all the other projects

PR #63, opened 2015-11-24
Fix errors in front50 caused by jamm not being a part of -javaagent

dpodder:
Commit 7ca8419 broke front50 in our testing, triggering the following exception every time we try to create an application:

java.lang.IllegalStateException: Instrumentation is not set; Jamm must be set as -javaagent

Looks like jvmArgs accepts an array instead of a string, so changing it to pass in a list of arguments rather than a single string fixed it for us.

### DIFF
--- a/front50-web/front50-web.gradle
+++ b/front50-web/front50-web.gradle
@@ -119,5 +119,5 @@ buildDeb {
 bootRun {
   // jamm is required when using prepared statements with in-memory C*
   def jamm = project.configurations.compile.find { it.name.startsWith('jamm') }
-  jvmArgs applicationDefaultJvmArgs.join(" ") + " -javaagent:${jamm}"
+  jvmArgs applicationDefaultJvmArgs + "-javaagent:${jamm}"
 }


dpodder:
@ewiseblatt This touches code you previously changed; could you check if it looks good?

ewiseblatt:
@tomaslin
Could you look at this? It doesnt look right to me, but I'm not that
familiar with these details.
Also, FYI daniel, there is a similar clause in echo.

-eric

dpodder:
Hi, Eric:

We haven't been using echo, so it's quite possible the issue exists there too and we just haven't hit it yet.

I'm new to Gradle so I'm certainly not an expert here either. Based on what I observed and the exception we were getting, the `-javaagent:<path_to_jamm>` was being ignored as an argument by the JVM. Some searching led me to [Gradle docs](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html#org.gradle.api.tasks.JavaExec:jvmArgs) suggesting that `jvmArgs` takes a `List<String>`--the implication being that it expects arguments to already be "tokenized," for lack of a better term. Keeping the arguments in array form and concatenating jamm to this array fixes the issue as far as I can tell.

tomaslin:
defer to @cfieber -

ewiseblatt:
@cfieber or @tomaslin

Could you have a look at this. The intent makes sense but I am not sure
whether this is syntactically correct
(' + "..."' vs '+ ["..."]'). How do I enable jamm to test it?

Note that echo has a similar gradle file so should change also.

- eric

tomaslin:
jamm is a static library used by cassandra, this works when you run via ./gradlew bootRun and can create an app.

PR #69, opened 2015-12-10
updates to latest kork-cassandra

cfieber:
keyspaces are now initialized in PostConstruct instead of on context refresh

Depends on spinnaker/kork#27

@ajordens ptal

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ allprojects { project ->
   group = "com.netflix.spinnaker.front50"
 
   spinnaker {
-    dependenciesVersion = "0.19.0"
+    dependenciesVersion = "0.25.0"
   }
   test {
     testLogging {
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -50,6 +54,9 @@ allprojects { project ->
     resolutionStrategy {
       force 'com.google.guava:guava:15.0'
       eachDependency {
+        if (it.requested.group == 'com.netflix.spinnaker.kork') {
```
cfieber: this needs to go before finally shipping it... yay library release train...

build.gradle
```diff
@@ -29,13 +29,17 @@ buildscript {
 }
 
 allprojects { project ->
+  repositories {
+    mavenLocal()
```
robfletcher: You sure you meant to commit that?

build.gradle
```diff
@@ -29,13 +29,17 @@ buildscript {
...
+  }
+
   apply plugin: 'spinnaker.project'
   apply plugin: 'groovy'
 
   group = "com.netflix.spinnaker.front50"
 
   spinnaker {
-    dependenciesVersion = "0.19.0"
+    dependenciesVersion = "0.23.0"
```
ajordens: Care to bump this up to 0.25.0.

build.gradle
ajordens: This can go once you've gone to 0.25.0

PR #73, opened 2016-03-14
Initial port of Front50 to S3

ajordens:
This PR introduces S3 as a supported persistence background in Front50.

Applications, Projects, Pipelines, Pipeline Strategies and Notifications are now stored as individual documents in S3.

Patterns:
- ${netflix.stack}/front50/applications
- ${netflix.stack}/front50/notifications
- ${netflix.stack}/front50/projects
- ${netflix.stack}/front50/pipelines
- ${netflix.stack}/front50/pipeline-strategies

To load data:
````
curl -X POST -H "Content-type: application/json" --data-binary @"notifications.json" http://localhost:7101/notifications/batchUpdate
curl -X POST -H "Content-type: application/json" --data-binary @"projects.json" http://localhost:7101/v2/projects/batchUpdate
curl -X POST -H "Content-type: application/json" --data-binary @"applications.json" http://localhost:7101/global/applications/batchUpdate
curl -X POST -H "Content-type: application/json" --data-binary @"pipelines.json" http://localhost:7101/pipelines/batchUpdate
curl -X POST -H "Content-type: application/json" --data-binary @"strategies.json" http://localhost:7101/strategies/batchUpdate
````

This PR does nothing to align APIs across notifications, projects, applications, pipelines or strategies (out of scope).

Disable Cassandra:
````
    cassandra
      enabled: false
````
Enable S3:
````
    spinnaker:
      s3:
        enabled: true
        bucket: __S3_BUCKET_NAME__
        rootFolder: ${netflix.stack}/front50/
````

Outstanding Items:
- front50-s3 tests are in-progress (I've successfully tested w/ gate+orca -> front50 -> s3)

### CODE REVIEW

File: build.gradle [hunk 1]
```diff
@@ -35,7 +35,7 @@ allprojects { project ->
   group = "com.netflix.spinnaker.front50"
 
   spinnaker {
-    dependenciesVersion = "0.25.0"
+    dependenciesVersion = "0.30.0"
```
ajordens: Included a kork-cassandra change allowing us to `cassandra.enabled: false`.

front50-cassandra/src/main/groovy/com/netflix/spinnaker/front50/model/application/CassandraApplicationDAO.groovy
```diff
@@ -85,18 +85,6 @@ class CassandraApplicationDAO implements ApplicationDAO {
              ) with compression={};''').execute()
     }
 
-    try {
-      // TODO: migrate schema on the fly, remove any time after 2/1/16.
```
ajordens: Long past 2/1/16.

front50-cassandra/src/main/groovy/com/netflix/spinnaker/front50/model/application/CassandraApplicationDAO.groovy
```diff
@@ -156,44 +144,12 @@ class CassandraApplicationDAO implements ApplicationDAO {
...
-          result = false
-        }
-      }
-      return result
-    } as Set
-
-    if (!items) {
-      throw new NotFoundException("No Application found for search criteria $attributes")
-    }
+    return ApplicationDAO.Searcher.search(all(), attributes);
```
ajordens: Split this logic out such that it could be similarly shared between Cassandra and S3.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/notification/HierarchicalLevel.groovy
```diff
@@ -13,14 +13,10 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.front50.notifications
+package com.netflix.spinnaker.front50.model.notification
 
 public enum HierarchicalLevel {
-    STAGE,
```
ajordens: AFAIK we are only using `APPLICATION` and `GLOBAL`.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/notification/Notification.groovy
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.notification
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.netflix.spinnaker.front50.model.Timestamped
+
+class Notification extends HashMap<String, Object> implements Timestamped {
```
ajordens: Notification/Pipeline/PipelineStrategy controllers were previously dealing in `Map's`. 

This is a bit of a hack but I needed to have all objects being stored in S3 implement a common interface.

front50-pipelines/src/main/groovy/com/netflix/spinnaker/front50/notifications/NotificationRepository.groovy
```diff
@@ -76,39 +80,27 @@ class NotificationRepository {
...
                 .getKey(CF_NAME)
                 .getColumn("$level:$name".toString())
                 .execute().result
-            notification = mapper.readValue(Zip.decompress(result.byteArrayValue), Map)
+            return mapper.readValue(Zip.decompress(result.byteArrayValue), Notification)
         } catch (NotFoundException ignored) {
-            notification = [email: []]
-        }
-        if (level == HierarchicalLevel.APPLICATION) {
-            NOTIFICATION_FORMATS.each {
```
ajordens: This add `GLOBAL` on top of `APPLICATION` logic has been moved into the Controller such that both S3 and Cassandra can benefit.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3NotificationDAO.java
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      S3Object s3Object = amazonS3.getObject(bucket, buildS3Key(name));
+      Notification notification = objectMapper.readValue(s3Object.getObjectContent(), Notification.class);
+      notification.setLastModified(s3Object.getObjectMetadata().getLastModified().getTime());
+      return notification;
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    } catch (AmazonS3Exception e) {
+      if (e.getStatusCode() == 404) {
+        return new Notification();
```
ajordens: Legacy behavior in that we expect an empty notification to be returned.

front50-cassandra/src/main/groovy/com/netflix/spinnaker/front50/model/application/CassandraApplicationDAO.groovy
```diff
@@ -41,7 +41,7 @@ import javax.annotation.PostConstruct
 
 @Slf4j
 @Component
-@ConditionalOnExpression('${spinnaker.cassandra.enabled:false}')
```
cfieber: worth checking the values for `cassandra.enabled` vs `spinnaker.cassandra.enabled` in spinnaker/spinnaker/config

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -0,0 +1,81 @@
...
+    return new RestTemplate();
+  }
+
+  @Bean
+  public ObjectMapper amazonObjectMapper() {
+    return new AmazonObjectMapper();
+  }
+
+  @Bean
+  @ConditionalOnExpression("${spinnaker.s3.enabled:false}")
```
cfieber: should this be just `s3.enabled` since the cassandra config went from `spinnaker.cassandra.enabled` to `cassandra.enabled`

ajordens:
There's a bit of boilerplate still between the various S3*DAOs ... I'm going to take another pass and see if I can clean it up further.

PTAL @spinnaker/netflix-reviewers @spinnaker/google-reviewers 

I've not had a chance to run this on a versioned bucket.

@tomaslin pointed out http://docs.basho.com/riakcs/latest/ which might be an interesting solution for anyone that can't run in AWS or against an S3-compatible API.

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
ajordens: We actually have config under `cassandra.*` and `spinnaker.cassandra.*`.

I suppose the breakdown is that cassandra.* comes from kork whereas spinnaker.cassandra.* is front50.

ajordens: Any strong preference @cfieber on spinnaker.s3 vs s3?

Part of me is inclined to drop the `spinnaker.`.

I could also updated CassandraConfig and add support for both `cassandra.cluster` and spinnaker.cassandra.cluster` values ... even though I suspect in the long run that Cassandra will be removed in entirety.

ewiseblatt: I prefer spinnaker.s3 to scope it.

I'd also add a corresponding entry to the front50.yml as well
spinnaker:
  s3:
     enabled: ${global.spinnaker.s3.enabled}

And add to spinnaker.yml (and default-spinnaker-local.yml)
global:
  spinnaker:
     s3:
       enabled: false

ewiseblatt: Make that
spinnaker:
  s3:
     enabled: ${global.spinnaker.s3.enabled:false}

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/application/ApplicationDAO.groovy
```diff
@@ -35,4 +35,46 @@ public interface ApplicationDAO {
 
   Set<Application> search(Map<String, String> attributes)
 
-}
\ No newline at end of file
+  void bulkImport(Collection<Application> applications)
+
+  static class Searcher {
```
ttomsu: This should probably be another interface, pluggable by cassandra, S3, or git

ttomsu: Nevermind - it doesn't look like there's anything cassandra specific about this.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/notification/NotificationDAO.groovy
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.notification
+
+interface NotificationDAO {
+  static final Collection<String> NOTIFICATION_FORMATS = ['sms', 'email', 'hipchat', 'slack']
```
ttomsu: Make this an enum?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
```diff
@@ -0,0 +1,60 @@
...
+    return super.get("id")
+  }
+
+  void setId(String id) {
+    super.put("id", id)
+  }
+
+  @Override
+  @JsonIgnore
+  Long getLastModified() {
```
ttomsu: This is a groovy file - just remove `private` from the field, no?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineStrategyDAO.groovy
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.pipeline
+
+interface PipelineStrategyDAO {
```
ttomsu: A lot of these DAOs are looking pretty similar. I imagine you could make one interface with java generics instead of enumerating out each interface (like https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ClusterProvider.groovy)

front50-pipelines/src/main/groovy/com/netflix/spinnaker/front50/notifications/NotificationRepository.groovy
```diff
@@ -57,8 +61,8 @@ class NotificationRepository {
         }
     }
 
-    List<Map> list() {
-        List<Map> notifications = []
+    List<Notification> list() {
```
ttomsu: Yay, real object types!!

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3NotificationDAO.java
ttomsu: Comments like these (any many other tidbits throughout this PR) should go in the code. If I were to change this code in the future, I wouldn't know to look for helpful information in Github.

ajordens: Fair enough, good point.

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
ajordens: The issue I suppose is that this is the only service with scoped config, which looks a little strange when I look at config on a service-by-service basis. 

And even then there will be core cassandra config expected to live at the root/unscoped level, ie:
`cassandra.embedded`.

Then again, there are plenty of inconsistencies in config styles across services so perhaps this is no worse.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/notification/NotificationDAO.groovy
ajordens: I just ripped it out of the Cassandra-specific `Repository`. 

I'm not sure it really adds much to be an `enum` except more LoC. 

TBH, Front50 is in an inconsistent state as far as coding styles go stemming back (_and possibly earlier_) to when 3 or 4 different services were squashed into one. 

If I had the time right now I'd nuke/deprecate all controllers and re-implement standard APIs on top of all the persistent entities ... but that will have to wait.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
ajordens: Hah ya, good catch.

That var used to be called something else before I renamed it to `lastModified`.

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
ewiseblatt: I see your point. Because it is s3, I was viewing that as a subsystem. Like
redis and cassandra. Unlike those, s3 isnt using the option, spinnaker is.
Also, in the presence of others, like gs or other platform variants, I'd
think you'd want them grouped/related together rather than in the global
space.
So the grouping category doesnt need to be 'spinnaker'. It could be
something else like dao. I dont have a feel for the big picture. I was
reacting to abstract policy more than the particular details.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineStrategyDAO.groovy
ajordens: Sure thing, had an earlier note to do another pass and cut out the boilerplate, I'll do that.

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
ajordens: I'm in agreement on grouping. 

Not a huge fan of `spinnaker` as the group (_don't feel it signifies anything_), but I'm inclined to leave it that way and avoid breaking existing configurations for now.

front50-test/src/main/groovy/com/netflix/spinnaker/front50/utils/CassandraTestHelper.groovy
```diff
@@ -0,0 +1,38 @@
...
+
+import com.netflix.astyanax.Keyspace
+import com.netflix.astyanax.connectionpool.impl.ConnectionPoolConfigurationImpl
+import com.netflix.spinnaker.kork.astyanax.AstyanaxComponents
+
+
+public class CassandraTestHelper {
+  AstyanaxComponents.EmbeddedCassandraRunner runner
+  Keyspace keyspace
+
```
ajordens: This is moved into `front50-test` and generified a bit such that it could be used in `front50-web` tests (see `PipelineControllerTCK` for an example).

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerTck.groovy
```diff
@@ -0,0 +1,185 @@
...
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.http.MediaType
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.setup.MockMvcBuilders
+
+abstract class PipelineControllerTck extends Specification {
```
ajordens: These tests used to be Mock-based which provided very little value.

I've started converting to be a lightweight TCK and test against `S3` and `Cassandra` DAOs.

Still some more tests that need flushing out but the confidence these will give us is significantly higher than anything else in Front50.

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/StrategyControllerTck.groovy
```diff
@@ -0,0 +1,188 @@
...
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.http.MediaType
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.setup.MockMvcBuilders
+
+abstract class StrategyControllerTck extends Specification {
```
ajordens: Sad face that we have a StrategyController and PipelineController and Pipeline/Strategy objects that are essentially identical. 

This is historical and not something I'll be able to change in this PR.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.groovy
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model
+
+import com.netflix.spinnaker.front50.exception.NotFoundException
+
+interface ItemDAO<T> {
```
ajordens: How's this look @ttomsu?

create()/update() seem a little strange to me but seem to fit at least the majority of the previous DAO implementations. 

I may still revisit the need for a `String id` parameter once I have the controller TCK tests completed.

ajordens:
The DAO refactor I just pushed up increased the size of this PR a fair bit.

I need to finish up the S3 and Cassandra tests but once that's done, I'm feeling reasonable confident that Front50 is in better shape than it was.

I owe some config updates to `spinnaker/spinnaker` once this is complete.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.groovy
ttomsu: LGTM

PR #76, opened 2016-03-18
Include each ItemDAO in the health check

ajordens:
- S3 just ensures it's refreshed itself in the past 45s
- Cassandra* executes a query

 * Only CassandraApplicationDAO executes a query, the other DAOs just return true (just tests Cassandra connectivity which should be the same regardless of DAO)


cfieber:
LGTM

PR #83, opened 2016-04-06
Upgrade to Spring Boot 1.2.8.RELEASE

gregturn:


cfieber:
This just changes the plugin version not any of the dependency versions. This update needs to flow thru kork and spinnaker-dependencies

PR #86, opened 2016-05-10
core: NPE thrown while reporting save error

lwander:
On application save: https://github.com/spinnaker/front50/blob/master/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/application/Application.groovy#L171 `originalApplication` is passed as `null`, causing the error reporting to fail.


duftler:
LGTM.

PR #89, opened 2016-05-12
History support

ajordens:
- Applications/Pipelines/Pipeline Strategies (to start)
- Only supported by S3 (Cassandra will return [])

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
@@ -22,9 +22,6 @@ import com.netflix.spinnaker.front50.model.Timestamped
 
 class Pipeline extends HashMap<String, Object> implements Timestamped {
   @JsonIgnore
-  Long lastModified
-
-  @JsonIgnore
   String getApplication() {
     return super.get("application")
   }
@@ -46,5 +43,17 @@ class Pipeline extends HashMap<String, Object> implements Timestamped {
 
   void setId(String id) {
     super.put("id", id)
+  }
+
+  @Override
+  @JsonIgnore
+  Long getLastModified() {
+    def updateTs = super.get("updateTs") as String
+    return updateTs ? Long.valueOf(updateTs) : null
+  }
+
+  @Override
+  void setLastModified(Long lastModified) {
+    super.put("updateTs", lastModified.toString())
   }
 }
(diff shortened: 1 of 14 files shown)


ajordens:
PTAL @spinnaker/netflix-reviewers 

A simple abstraction over going straight to an S3 bucket to look at historical pipeline/strategy/application configurations.

Currently just returns a full set of objects with timestamps ... figure that any `diff` should occur upstream.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
```diff
@@ -22,9 +22,6 @@ import com.netflix.spinnaker.front50.model.Timestamped
 
 class Pipeline extends HashMap<String, Object> implements Timestamped {
   @JsonIgnore
-  Long lastModified
```
ajordens: I don't believe having this actually go across the wire will cause grief but I'll give it a whirl in prestaging.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineDAO.groovy
```diff
@@ -23,4 +23,5 @@ import com.netflix.spinnaker.front50.model.ItemDAO
 interface PipelineDAO extends ItemDAO<Pipeline> {
   String getPipelineId(String application, String pipelineName)
   Collection<Pipeline> getPipelinesByApplication(String application)
+  Collection<Pipeline> getPipelineHistory(String id)
```
cfieber: should these be bounded somehow (max results param or something)?

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3Support.java
```diff
@@ -113,6 +113,30 @@ public T findById(String id) throws NotFoundException {
     }
   }
 
+  public Collection<T> allVersionsOf(String id) throws NotFoundException {
+    try {
+      VersionListing versionListing = amazonS3.listVersions(new ListVersionsRequest(bucket, buildS3Key(id), null, null, null, 20));
```
cfieber: ah - I see 20 maxResults here - maybe make that configurable (and / or support it as a request option)

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineDAO.groovy
ajordens: Probably ... I just shoved in an explicit max of 20.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
```diff
@@ -48,6 +48,15 @@ class StrategyController {
         pipelineStrategyDAO.getPipelinesByApplication(application)
     }
 
+    @RequestMapping(value = '{id:.+}/history', method = RequestMethod.GET)
+    Collection<Pipeline> getHistory(@PathVariable String id) {
+        try {
+            return pipelineStrategyDAO.getPipelineHistory(id)
+        } catch (UnsupportedOperationException ignored) {
+            return []
```
cfieber: if the operation is supported, is the current version one of the values returned?

if it is, should the unsupported case just return the current value, or should the history request filter out the current value?

cfieber:
LGTM

PR #90, opened 2016-05-14
Initial pass at using Google Cloud Storage instead of cassandra.

ewiseblatt:
@duftler, @ajordens 

This doesnt have tests yet, but seems to work when I run it.
I'm not sure how to port the test because it is a live test (not a mock) so needs credentials to run.

Adam,
I'd like to refactor this to remove a lot of the copy/paste stuff.
I'm thinking about changinging the Support class to a templated base class, removing the actual storage mechanism from it, and injecting that back in with an interface. If that's possible, then it will concentrate the mechanism in one place and clarify what it needs to do and will make the behavior stuff more consistent and less brittle across different implementations. I'd move the feature implementation classes into front50-core (?) but and just keep the mechanism class and config class in the front50-<mechanism> components.

Does that make sense? I could do that after this PR.

### DIFF
--- a/front50-gcs/front50-gcs.gradle
+++ b/front50-gcs/front50-gcs.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2016 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,17 +14,14 @@
  * limitations under the License.
  */
 
-rootProject.name = "front50"
+dependencies {
+  compile project(":front50-core")
+  compile 'io.reactivex:rxjava:1.0.10'
 
-include 'front50-web', 'front50-core', 'front50-cassandra', 'front50-gcs', 'front50-pipelines', 'front50-redis', 'front50-s3', 'front50-test'
+  spinnaker.group('google')
+  /* spinnaker.group('googleStorage') */
+  compile 'com.google.apis:google-api-services-storage:v1-rev72-1.22.0'
 
-def setBuildFile(project) {
-  project.buildFileName = "${project.name}.gradle"
-  project.children.each {
-    setBuildFile(it)
-  }
-}
 
-rootProject.children.each {
-  setBuildFile(it)
+  testCompile project(":front50-test")
 }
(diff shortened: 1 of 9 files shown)


ajordens:
Out of curiosity, did you try using an `AmazonS3Client` as is and just replace the `endpoint` URL and authentication keys?

In test cases using [s3proxy](https://github.com/andrewgaul/s3proxy), I did the following:

```
// See `S3ApplicationDAOSpec` and `S3TestHelper.s3ProxyUnavailable()`.
// I couldn't find a way to programatically spin up the s3proxy so these tests only run if someone first starts the proxy
def amazonS3 = new AmazonS3Client(new ClientConfiguration())
amazonS3.setEndpoint("http://127.0.0.1:9999")
```

As for authentication, you [should](https://cloud.google.com/storage/docs/migrating#authentication) be able to supply the GCS access and secret keys in lieu of the AWS access and secret equivalents. Not sure if the `oauth2` credentials you're using in here are compatible.

FWIW `S3Support` initially started out as a wrapper around the periodic refresh behavior. This was necessary primarily because S3 was slow to return the contents of buckets with ~2k things in them. It's specifically optimized to make as few calls as possible to S3.

Maybe GCS isn't and the complexity of refreshing a cache would not be worth it.

ewiseblatt:
On Sat, May 14, 2016 at 11:36 AM, Adam Jordens <notifications@github.com>
wrote:

> Out of curiosity, did you try using an AmazonS3Client as is and just
> replace the endpoint URL and authentication keys?
>
> In test cases using s3proxy <https://github.com/andrewgaul/s3proxy>, I
> did the following:
>
Interesting, but I dont understand how to use it. The docs talk about the
endpoint in terms of a host and port, not a base URL (GCS is
https://www.googleapis.com/storage/v1) Credentials are key/secret but we
use service accounts, which I think are a bit different.
And looking it over, isnt clear whether this works or is an interface I
need to implement. The GCS docs suggest that I can send S3 requests and
just use google credentials. But looking at the AWS java API has the same
issues; It isnt clear where or how to change the endpoint and credentials
dont seem to support our service accounts. I could be wrong, but that's my
impression looking at it. I'll ask around.

ajordens:
```
def amazonS3 = new AmazonS3Client(new BasicAWSCredentials("ACCESS_KEY", "SECRET_KEY"), new ClientConfiguration())
amazonS3.setEndpoint("http://127.0.0.1:9999")
```

It still might not support a service account ... but sure would be nice if you could get credentials dedicated to it and try and use as-is ... without refactoring.

PR #93, opened 2016-05-19
Added Google Cloud Storage alternative to Cassandra.

ewiseblatt:
@duftler, @ajordens 

I'm not going to refactor.

Also, it turns out you cannot batch operations that upload/download storage content.
That's too bad, since the content is small and the size of the JSON data that would otherwise be in a batched query/response.

### DIFF
--- a/front50-gcs/front50-gcs.gradle
+++ b/front50-gcs/front50-gcs.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2016 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,17 +14,14 @@
  * limitations under the License.
  */
 
-rootProject.name = "front50"
+dependencies {
+  compile project(":front50-core")
+  compile 'io.reactivex:rxjava:1.0.10'
 
-include 'front50-web', 'front50-core', 'front50-cassandra', 'front50-gcs', 'front50-pipelines', 'front50-redis', 'front50-s3', 'front50-test'
+  spinnaker.group('google')
+  /* spinnaker.group('googleStorage') */
+  compile 'com.google.apis:google-api-services-storage:v1-rev72-1.22.0'
 
-def setBuildFile(project) {
-  project.buildFileName = "${project.name}.gradle"
-  project.children.each {
-    setBuildFile(it)
-  }
-}
 
-rootProject.children.each {
-  setBuildFile(it)
+  testCompile project(":front50-test")
 }
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.front50.config;
+
+import com.netflix.spinnaker.front50.model.*;
+import com.google.api.services.storage.Storage;
+import com.google.api.services.storage.StorageScopes;
+import com.google.api.services.storage.model.Bucket;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.HttpResponseException;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import rx.schedulers.Schedulers;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.concurrent.Executors;
+import java.util.Collections;
+
+@Configuration
+public class GcsConfig {
+  // Refresh every 10 minutes. In practice this either doesnt matter because refreshes are fast enough,
+  // or should be finer tuned. But it seems silly to refresh at a fast rate when changes are generally infrequent.
+  // Actual queries always check to see if the cache is out of date anyway. So this is mostly for the benefit of
+  // keeping other replicas up to date so that last-minute updates have fewer changes in them.
+  private static int APPLICATION_REFRESH_INTERVAL = 10 * 60;
+  private static int PROJECT_REFRESH_INTERVAL = 10 * 60;
+  private static int NOTIFICATION_REFRESH_INTERVAL = 10 * 60;
+  private static int PIPELINE_REFRESH_INTERVAL = 10 * 60;
+  private static int PIPELINE_STRATEGY_REFRESH_INTERVAL = 10 * 60;
+
+  @Value("${spinnaker.gcs.bucket}")
+  private String bucket;
+
+  @Value("${spinnaker.gcs.rootFolder}")
+  private String rootFolder;
+
+  @Value("${providers.google.primaryCredentials.jsonPath}")
+  private String jsonPath;
+
+  @Value("${providers.google.primaryCredentials.project}")
+  private String project;
+
+  @Value("${Implementation-Version:Unknown}")
+  private String applicationVersion;
+
+  private void ensureBucket(Storage storage) throws IOException {
+    Logger log = LoggerFactory.getLogger(getClass());
+    try {
+        Bucket bkt = storage.buckets().get(bucket).execute();
+        boolean has_versioning = bkt.getVersioning().getEnabled();
+        log.info("Bucket versioning is {}.",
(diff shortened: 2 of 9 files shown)


front50-gcs/front50-gcs.gradle
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compile project(":front50-core")
+  compile 'io.reactivex:rxjava:1.0.10'
+
+  spinnaker.group('google')
+  /* spinnaker.group('googleStorage') */
```
ajordens: Can you remove the commented out `googleStorage`, assuming it's not necessary.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+  private String jsonPath;
+
+  @Value("${providers.google.primaryCredentials.project}")
+  private String project;
+
+  @Value("${Implementation-Version:Unknown}")
+  private String applicationVersion;
+
+  private void ensureBucket(Storage storage) throws IOException {
+    Logger log = LoggerFactory.getLogger(getClass());
```
ajordens: The `Logger` should just be defined statically on the class

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+  private static int PIPELINE_REFRESH_INTERVAL = 10 * 60;
+  private static int PIPELINE_STRATEGY_REFRESH_INTERVAL = 10 * 60;
+
+  @Value("${spinnaker.gcs.bucket}")
+  private String bucket;
+
+  @Value("${spinnaker.gcs.rootFolder}")
+  private String rootFolder;
+
+  @Value("${providers.google.primaryCredentials.jsonPath}")
```
ajordens: `providers.google` seems like an odd configuration path ... certainly for `front50`. Is this config path used in other services?

Why wouldn't you keep this under `spinnaker.gcs`?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+  private String jsonPath;
+
+  @Value("${providers.google.primaryCredentials.project}")
+  private String project;
+
+  @Value("${Implementation-Version:Unknown}")
+  private String applicationVersion;
```
ajordens: What is this used for? Not sure where `Implementation-Version` would be getting set?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+    try {
+        Bucket bkt = storage.buckets().get(bucket).execute();
+        boolean has_versioning = bkt.getVersioning().getEnabled();
+        log.info("Bucket versioning is {}.",
+                 has_versioning ? "enabled" : "DISABLED");
+    } catch (HttpResponseException e) {
+      if (e.getStatusCode() == 404) {
+          log.warn("Bucket {} does not exist. Create it in project={}",
+                   bucket, project);
+          Bucket.Versioning versioning = new Bucket.Versioning().setEnabled(true);
```
ajordens: I know in S3, it's not necessary common that everyone would have permission to create a bucket.

Is it always the case if you're running with a service account?

This was primarily the reason why we didn't automatically create the bucket as part of initialization on the S3-side.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+          Bucket spec = new Bucket().setName(bucket).setVersioning(versioning);
+          try {
+              storage.buckets().insert(project, spec).execute();
+          } catch (IOException e2) {
+              log.error("Could not create bucket={} in project={}: {}",
```
ajordens: Are these not legitimate exception cases if the bucket cannot be created ... seems like an exception should be thrown.

Otherwise you're going to get a `Storage` client out of `googleCloudStorageClient()` that would never actually work?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.concurrent.Executors;
+import java.util.Collections;
+
+@Configuration
+public class GcsConfig {
+  // Refresh every 10 minutes. In practice this either doesnt matter because refreshes are fast enough,
+  // or should be finer tuned. But it seems silly to refresh at a fast rate when changes are generally infrequent.
+  // Actual queries always check to see if the cache is out of date anyway. So this is mostly for the benefit of
+  // keeping other replicas up to date so that last-minute updates have fewer changes in them.
+  private static int APPLICATION_REFRESH_INTERVAL = 10 * 60;
```
ajordens: If you're going to to point of declaring these variables ... should use `TimeUnit.MINUTES.toMillis(10)`, etc. and probably name the variables with a _MS suffix for consistency.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -0,0 +1,174 @@
...
+    return new RestTemplate();
+  }
+
+  @Bean
+  public ObjectMapper gcsObjectMapper() {
+    return new ObjectMapper();
+  }
+
+  @Bean
+  @ConditionalOnExpression("${spinnaker.gcs.enabled:false}")
```
ajordens: Probably worth while to just put this `@ConditionalOnExpression` at the class-level.

Otherwise you're going to get some beans created when GCS is not enabled.

ajordens:
I have a few concerns around the copy and pasted DAO and Support class for an otherwise nearly identical implementation.

If we really cannot use/expect a user to be able to fetch ACCESS_KEY/SECRET_KEY (with which you shouldn't need to do anything but instantiate the `AmazonS3Client`), than I would suggest we introduce an interface that supports:

```
getObject(bucketName, objectKey)
getObject(bucketName, objectKey, versionId)
listObjects(bucketName, folderName)
listVersions(bucketName, objectKey)
putObject(bucketName, objectKey, object)
```

Far less code overall I suspect. 

Thoughts?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ewiseblatt: I'm not exactly sure where this is set, but it appears to be a standard Java convention if not provided by JDK itself. This is used to set the version in the UserAgent so that it is consistent with the debian package version as well.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ewiseblatt: Static or instance?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ewiseblatt: I think this is currently the case because security is coarse, but will eventually not be the case (as is with you). However the service account might still have permissions. It isnt to say that it has to work this way, but wanted to reduce the amount of first-time configiguration if I could. If you are running from within the same project, you have full permissions anyway so why not try.

front50-gcs/front50-gcs.gradle
ewiseblatt: I am going to add this to spinnaker-dependencies

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -41,40 +41,44 @@
...
   // Actual queries always check to see if the cache is out of date anyway. So this is mostly for the benefit of
   // keeping other replicas up to date so that last-minute updates have fewer changes in them.
-  private static int APPLICATION_REFRESH_INTERVAL = 10 * 60;
-  private static int PROJECT_REFRESH_INTERVAL = 10 * 60;
-  private static int NOTIFICATION_REFRESH_INTERVAL = 10 * 60;
-  private static int PIPELINE_REFRESH_INTERVAL = 10 * 60;
-  private static int PIPELINE_STRATEGY_REFRESH_INTERVAL = 10 * 60;
+  private final Logger log = LoggerFactory.getLogger(getClass());
+  private static int APPLICATION_REFRESH_MS = (int)TimeUnit.MINUTES.toMillis(10);
+  private static int PROJECT_REFRESH_MS = (int)TimeUnit.MINUTES.toMillis(10);
```
ajordens: Just use a `long` to avoid the cast.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -145,30 +148,30 @@ public ObjectMapper gcsObjectMapper() {
   @Bean
   @ConditionalOnExpression("${spinnaker.gcs.enabled:false}")
```
ajordens: You don't need these `@ConditionalOnExpression`

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -41,40 +41,44 @@
...
   private String project;
 
   @Value("${Implementation-Version:Unknown}")
   private String applicationVersion;
 
   private void ensureBucket(Storage storage) throws IOException {
-    Logger log = LoggerFactory.getLogger(getClass());
     try {
         Bucket bkt = storage.buckets().get(bucket).execute();
         boolean has_versioning = bkt.getVersioning().getEnabled();
```
ajordens: Minor but camel case

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ewiseblatt: The problem is that the use site is an int. I thought it made more sense to cast here both because it is simpler and clear the magnitude fits.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ewiseblatt: I was trying to be consistent with S3 for this initial stab. Since you didnt check for the bucket existing (and bucket could be deleted at some point after anyway) I didnt want to prevent a startup and let the error handling on save worry about it. I dont mind putting an exception here.

ewiseblatt:
Closing this in favor of the refactoring variant

PR #94, opened 2016-05-26
Added Google Cloud Storage alternative to Cassandra.

ewiseblatt:
@ajordens 
This is another take on the previous GCS PR but assumes a refactoring.
I put the refactoring in the google-gcs package here, because I'm not sure where to put it and want to run it by you before pursuing it further. I left S3 alone, but it is based on what you did so should be straight forward to refactor.

The design uses a base BucketDAO class with the item subclasses as before, except that you inject a StorageService into the BucketDAO constructor where the StorageService is specialized for the underlying service implementation.

In the BucketDAO I cut out this block https://github.com/spinnaker/front50/blob/master/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3Support.java#L256
because it seemed superfluous. I suspect it really is not and there is an extra bit of concurrency in there. If that is the case, then I need to add another method into the StorageService to get at the metadata and another one to load an object from a metadata specification. If that is the case then the metadata type would be another class parameter to the StorageService and maybe there is a marker interface or something to pass through the BucketDAO so it can mediate between the two StorageService APIs.

The interface doesnt really support versioning because I dont think the existing data model supports it. I have some comments in there about how it could be moved forward, but doing so is outside the scope of this PR.

Note that to keep it simpler I'm using the same filename for all the types, but restricting search on the path that branches at the root depending on the type. Since I assume you want to preserve compatability with what you've already deployed, I was imaging that the S3StorageService would map from the daoTypeName into the file name to use, and encapsulate that mapping. If you add new types, then you could either extend the mapping, or default to some standard filename.

### DIFF
--- a/front50-gcs/front50-gcs.gradle
+++ b/front50-gcs/front50-gcs.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2016 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,17 +14,14 @@
  * limitations under the License.
  */
 
-rootProject.name = "front50"
+dependencies {
+  compile project(":front50-core")
+  compile 'io.reactivex:rxjava:1.0.10'
 
-include 'front50-web', 'front50-core', 'front50-cassandra', 'front50-gcs', 'front50-pipelines', 'front50-redis', 'front50-s3', 'front50-test'
+  spinnaker.group('google')
+  /* spinnaker.group('googleStorage') */
+  compile 'com.google.apis:google-api-services-storage:v1-rev72-1.22.0'
 
-def setBuildFile(project) {
-  project.buildFileName = "${project.name}.gradle"
-  project.children.each {
-    setBuildFile(it)
-  }
-}
 
-rootProject.children.each {
-  setBuildFile(it)
+  testCompile project(":front50-test")
 }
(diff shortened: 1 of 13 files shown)


ajordens:
On my phone but at first glance this looks good!

I'll give it a proper look through tomorrow but I'd be inclined to get it
merged down and I can do the work for S3.

ewiseblatt:
I added a test based on the current behavior, which is consistent with the s3 reference implementation.

Note that it doesnt update the cache in places you might expect it to since it relies on future querying of the timestamps. I'm wondering if the timestamp management should be moved up into the BucketDAO rather than relying on the StorageService. Either way there seems to be race conditions but I'm not sure how significant they are -- just missing isnt so much a concern to me since it could be written off as poor timing. However if the timestamp in the bucket is older than it should be then you are going to lose updates until it is written again or restarted. It isnt clear to me if this happens or not and the API is complicated to test this type of granularity (or at least I'm not experienced with Java or this style).

I'm not sure why the startRefresh test doesnt cause an update. I'm still looking into fixing that. I cannot tell if it is just latency because of another thread (which I tried to accomdate with a Sleep) , or there are components not initialized properly in the test. It seems to behave properly in the production runtime, though not sure about the latency. I wanted to mock it out, but the black-box scheduler interactions are too tedious and brittle, and I dont yet understand the API to write a fake. I'm was hoping to add a quick cycle time for the test in this PR and put off actually testing the scheduler and concurrency for another time (or ask someone else who is familiar with those APIs)

ajordens:
Please squash and otherwise this is fine to merge.

I'll find some time next week to do a bit of refactoring and make s3 work with it.

PR #96, opened 2016-06-10
validate pipelines have applications and names

robfletcher:
Saving a pipeline with no `application` property means we can never retrieve any pipelines until the database gets cleaned up. This is quite bad.

PTAL @ajordens


ajordens:
LGTM

PR #98, opened 2016-06-20
Added web interfaces getById and getByName for pipeline and strategy

ewiseblatt:
@ajordens This is the front50 fix for the issue I raised last week.

I dont know where the consumers for this would be. I'd like deck to ask for pipelines similar to applications so that it hits the storage service to check cache freshness. I assume orca would want to do something similar as well.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
@@ -44,6 +44,18 @@ class PipelineController {
     pipelineDAO.getPipelinesByApplication(application)
   }
 
+  @RequestMapping(value = 'getById/{id:.+}', method = RequestMethod.GET)
+  Pipeline findById(@PathVariable String id) {
+    pipelineDAO.findById(id)
+  }
+
+  @RequestMapping(value = 'getByName/{application:.+}/{name:.+}', method = RequestMethod.GET)
+  Pipeline findByName(@PathVariable String application, @PathVariable String name) {
+    pipelineDAO.findById(
+            pipelineDAO.getPipelineId(application, name)
+    )
+  }
+
   @RequestMapping(value = '{id:.+}/history', method = RequestMethod.GET)
   Collection<Pipeline> getHistory(@PathVariable String id,
                                   @RequestParam(value = "limit", defaultValue = "20") int limit) {
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
@@ -49,6 +49,18 @@ class StrategyController {
         pipelineStrategyDAO.getPipelinesByApplication(application)
     }
 
+    @RequestMapping(value = 'getById/{id:.+}', method = RequestMethod.GET)
+    Pipeline findById(@PathVariable String id) {
+        pipelineStrategyDAO.findById(id)
+    }
+
+    @RequestMapping(value = 'getByName/{application:.+}/{strategy:.+}', method = RequestMethod.GET)
+    Pipeline findByName(@PathVariable String application, @PathVariable String strategy) {
+        pipelineStrategyDAO.findById(
+                pipelineStrategyDAO.getPipelineId(application, strategy)
+        )
+    }
+
     @RequestMapping(value = '{id:.+}/history', method = RequestMethod.GET)
     Collection<Pipeline> getHistory(@PathVariable String id,
                                     @RequestParam(value = "limit", defaultValue = "20") int limit) {


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -44,6 +44,18 @@ class PipelineController {
     pipelineDAO.getPipelinesByApplication(application)
   }
 
+  @RequestMapping(value = 'getById/{id:.+}', method = RequestMethod.GET)
```
ajordens: I don't believe we should be explicitly encoding verbs (like `getById`/`getByName`) in query paths ... I would also drop the lookup by name and only support `id`. These are new API endpoints without callers, we don't need to support lookups by the deprecated name.

The API around pipelines is rather inconsistent in that we have `/pipelines/{application}' where it should be `/pipelines/{id}` IMO. The ability to lookup by application would make more sense if `application` where a query parameter on `/pipelines`.

Given that's where we are right now, I would lean towards `/pipelines/{application}/{id}`.

ewiseblatt: I agree on the query path. However I think I need to add a new prefix because otherwise I think the URLs are ambiguous.
GET /pipelines/{application}
GET /pipelines/{id}/history

Isnt GET/pipelines/{application}/{id} ambiguous with history where id=history? I know that history inst a valid id, but does the matching mechanism know that?

I went with getById thinking I was consistent with the DELETE /pipelines/deleteById.

There is a DELETE /pipelines/{application}/{pipeline}. Ideally the GET would match that.
Can we move the history URL or is that in active use?

ajordens: We can totally move that history URL ... /pipelines/{application}/{id}/history would make sense I suppose.

That `deleteById` is equally bad (_but since day one I suspect_) ... it'd be great to have DELETE /pipelines/{application}/{id}.

ewiseblatt: ok. I need to create a new PR unless you know how I can checkout a branch with a slash. I didnt realize I'd hit this problem when I created the branch. Stackoverflow suggests it should be working anyway, but git seems to be interpreting the branch as a path. I'm going to give up and create a new PR.

Why are you using {application}/{id}? Arent ID's global?

There is already DELETE /pipelines/{application}/{pipeline}. Wouldnt .../{id} be redundant to that?
I thought I saw somewhere that tried to treat the value as an id and if that fails to treat it as a name. I dont recall where I saw that. I didnt like that, but if you are looking to transition from using names to using ids then that would be one way to do it. Since the IDs are internal and the names are external, I think having interfaces by name (at least variants) makes sense.

ajordens: I believe the API should be similar to:

```
GET /pipelines/{id} 
GET /pipelines/{id}/history
DELETE /pipelines/{id} 
PUT /pipelines/{id} 

POST /pipelines
GET /pipelines
GET /pipelines?name=abc
GET /pipelines?application=xyz
```

In the previous comment I included {application} because /pipeline/{id} is overloaded to the extent of not being usable (_if it's an application, we expect a collection ... if it's an id, we expect a single result_).

We could bite bullet and version since what is being talked about here is a somewhat significant change (/v2/pipelines?) affecting multiple clients.

ewiseblatt: I dont want a v2 or client impact in this CL. I'd like to converge S3 and
GCS first, and I'm leaving for vacation next week.
The proposal here is enabling future changes while still supporting current
usage patterns.

I am interpreting your previous reply (and out of band discussion earlier
this morning) as the following:

Current API

global scope means multiple instances across multiple apps
application scope means multiple instances in app, instance = one instance)
instance scope means one specific instance


global list       GET /pipelines
global save       POST /pipelines
global update     POST /pipelines/batchUpdate
instance delete   DELETE /pipelines/deleteById/{id}
instance delete   DELETE /pipelines/{application}/{pipeline}
instance rename   POST /pipelines/move

instance history  GET  /pipelines/{id}/history
application list  GET /pipelines/{application}


Proposed API (in this PR)

DEPRECATE (keep around but discourage use)
  instance delete   DELETE /pipelines/deleteById/{id}
 instance history  GET  /pipelines/{id}/history
 application list  GET /pipelines/{application}

ADD  (in this PR)
   instance update   PUT /pipelines/{id}
 instance get      GET /pipelines?id={id}
 application list  GET /pipelines?application={application}

And in addition Consider Extending the API  (in future PR)

  instance version-meta   GET /pipelines/versions/{id}
  instance version        GET /pipelines?id={id}&version={version}
  instance restore        PUT /pipelines/restore/{id}/{version}
                                                 or    PUT
/pipelines/{id}?version={version}
                                                 but this doesnt take a
payload like update does so might be dangerous.
    application  some way to list deleted pipeline ids



It would be nice if Applications (and notifications) were consistent with
these. Notifications look a bit fishy so am putting that off for now (but
raising them since I'm asking you to confirm how this can end up
consistent, which will need to accomodate them).
Applications have .../applications/name/{application} for methods that
operate on instances where .../applications and
.../applications/batchUpdate operate in the equivalent of global. If that
is the case, I propose pipelines use /id/ for the instance APIs.

ADD  (in this PR)
  instance update   PUT /pipelines/id/{id}
 instance get      GET /pipelines/id/{id}
 application list  GET /pipelines/application/{application}

This conflicts with  the existing GET /pipelines/{id}/history. Until it can
be removed, the implementations of the ambiguous APIs can check their
parameters for "id" or "history" depending on which, and call the right
handling. Unless actually specifying this causes a error from the ambiguous
specification.

The proposed extension would also use the /id/ paths, except for restore.



Thoughts?

ewiseblatt: Actually instead of /id/, should be /pipeline/ or /instance/

ewiseblatt:
So, it turns out that there is no static/runtime ambiguity raised between {id}/history and instance/{id}.
So I can just add an ID getter (and PUT) in, or I can add deprecations and alternate APIs to make consistent naming.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
ajordens: I'm not really following what is proposed.

I wouldn't want to see `/pipelines/id/{id}` or `/pipelines/application/{application}` or `/pipelines?id=xxx`.

Given that all these are additional, there's no real harm in:
```
/v2/pipelines/{id}
/v2/pipelines?name=xxx
```

You only really want to add the ability to fetch a pipeline by id or name ... and this seems the simplest thing to do. You would need to track down callers that should be using the fetch by id anyways.

ewiseblatt: Ok. Reset. A couple questions.
/v2/pipelines would add a new controller. Is that ok? Swagger wants them to have different names (even though they are in different packages). I called it a V2PipelineController. Now the V2 controller has a couple methods and the old controller has the rest. Its fine functionally, but the API is odd, at least when going through the swagger web ui. I dont have stake in that so dont care, but want to verify.

Also, "name" is the pipeline name. Names are scoped to applications. So if you are providing a name, dont you also need the application?

ajordens: Yup, I think that's fine on both counts (having a v2 and non-v2
api/controller).

Eventually the v2 api will be updated to include the missing functionality
from v1, and v1 becomes deprecated. Front50 is an internal API and I
wouldn't expect direct calls to it from anything but other Spinnaker
services (it should either require two-way ssl or be security
group/firewalled).

You're absolutely correct on the last part, would need to supply both
application and name query parameters. I wouldn't have a problem if the API
supported application, name, application + name or only application,
application + name (and raised a non-200 if only name were provided). In
the former, if you just supplied name, you would get the similarly named
pipelines across applications. Not particularly common.

Doing this is a nice step forward towards a more sensible API and the
patterns can be applied to the other types (applications, notifications,
etc.).

ajordens:
Going to close this PR ... it's hung around for a few months and afaik isn't a particular priority for either of us right now.

PR #102, opened 2016-07-07
optionally specify region on AmazonS3 client

cfieber:
without region set S3 requests get a redirect from the generic s3.amazonaws.com to the appropriate regional endpoint. this works but it's an extra hop


ajordens:
Nice.

LGTM

PR #103, opened 2016-07-14
run front50 with S3 backend behind a corporate proxy

janwillies:
To use edit config/spinnaker-local.yml:
- storage_bucket: YOUR_S3_BUCKET
- bucket_root: front50
- region: eu-central-1
- proxyHost: webproxy.foo.bar
- proxyPort: 8080
- proxyProtocol: HTTP

Currently only works with proxies without credentials


ajordens:
Sorry I was OOO for the past week.

This looks good to me, do you mind rebasing and I'll get it merged down.

Thanks!

PR #104, opened 2016-07-18
Adds ApplicationPermission model object, DAO for GCS/S3, and controller.

ttomsu:
New endpoint will be used by Fiat when starting up to load applications and periodically update its cache.  POST/PUT endpoints can be used by Orca when new applications are created.

Deprecates CassandraDAO since NFLX is no longer using it and we're encouraging new customers to go the GCS/S3 bucket route.

@ajordens @cfieber @jtk54 @duftler PTAL.

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
@@ -17,15 +17,15 @@
 package com.netflix.spinnaker.front50.config;
 
 import com.netflix.spinnaker.front50.model.*;
-
+import com.netflix.spinnaker.front50.model.application.ApplicationPermissionDAO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.client.RestTemplate;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -61,9 +61,17 @@
...
   @Bean
-  public GcsStorageService googleCloudStorageService() {
+  public GcsStorageService defaultGoogleCloudStorageService() {
+    return googleCloudStorageService(null /*dataFilename*/);
+  }
+
+  private GcsStorageService googleCloudStorageService(String dataFilename) {
     GcsStorageService service;
-    service = new GcsStorageService(bucket, rootFolder, project, jsonPath, applicationVersion);
+    if (dataFilename == null) {
```
jtk54: Why not `!dataFilename` here? Seems like you would want the first case if it's an empty string as well.

ttomsu: Forget it, Jake; it's Javatown.

jtk54: Ah, yes, we meet again.

jtk54:
LGTM, giving others a chance to review.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
duftler: Should you also be checking `dataFilename.empty()`?

duftler:
LGTM. Defer to @cfieber on merging.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ttomsu: Done.

cfieber:
LGTM too

PR #105, opened 2016-07-20
Adds Service Account model object, DAOs, and controller.

ttomsu:
Moving static configuration of service accounts from Fiat to Front50.

@cfieber @duftler @jtk54 PTAL

@ajordens FYI


jtk54:
LGTM, deferring to others to merge.

PR #106, opened 2016-07-20
Adds ability to specify location of GCS auto-created bucket.

ttomsu:
Fixes https://github.com/spinnaker/spinnaker/issues/1002. Buckets already created cannot have their location changed - you must move the data, delete it and have Front50 recreate the bucket. See https://cloud.google.com/storage/docs/managing-buckets#manage-class-location

@ewiseblatt @duftler @lwander PTAL

CC: @leelasharma

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
@@ -48,6 +48,9 @@ public class GcsConfig {
   @Value("${spinnaker.gcs.bucket}")
   private String bucket;
 
+  @Value("${spinnaker.gcs.bucketLocation}")
+  private String bucketLocation;
+
   @Value("${spinnaker.gcs.rootFolder}")
   private String rootFolder;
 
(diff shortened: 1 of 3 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -48,6 +48,9 @@
   @Value("${spinnaker.gcs.bucket}")
   private String bucket;
 
+  @Value("${spinnaker.gcs.bucketLocation}")
```
ewiseblatt: I think you should add an entry to front50.yml (probably pulling the value from spinnaker.yml) and link a comment to
https://cloud.google.com/storage/docs/bucket-locations as a means to document this config parameter.

You dont need to bother with it in default-spinnaker-local.yml .

The default value could be empty.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -82,20 +84,27 @@ private GoogleCredential loadCredential(
   }
 
   @VisibleForTesting
-  GcsStorageService(String bucketName, String basePath, String projectName, Storage storage) {
+  GcsStorageService(String bucketName,
+                    String bucketLocation,
```
ewiseblatt: Really this is defaultBucketLocation. It doesnt ensure the bucket is here, rather it is the default place it will create them. The potential confusion is that this has no effect on existing buckets. A comment is sufficient if you want to keep the name.

ewiseblatt:
LGTM after comments

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
ttomsu: Roger - I'll put that in in the spinnaker/spinnaker repo.

PR #111, opened 2016-08-01
Generic migration support

ajordens:
- An initial migrator that will convert linear pipelines to parallel (valid until 2016-09-01)

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/LinearToParallelMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/LinearToParallelMigration.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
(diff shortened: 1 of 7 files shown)


ajordens:
FYI @spinnaker/google-reviewers

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/LinearToParallelMigration.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.lang.String.format;
+
+@Component
+public class LinearToParallelMigration implements Migration {
+  private static final Logger log = LoggerFactory.getLogger(LinearToParallelMigration.class);
+
+  // Only valid until September 1st, 2016
+  private static final Date VALID_UNTIL = new GregorianCalendar(2016, 8, 1).getTime();
```
robfletcher: java.time is your friend

PR #113, opened 2016-08-10
Add timestamp marker files to avoid future 404 error noise

ewiseblatt:
Also attempt an update before an insert since it is far less likely to fail.

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
@@ -27,6 +27,7 @@ import com.google.api.services.storage.model.Bucket;
 import com.google.api.client.http.ByteArrayContent;
 import com.google.api.client.util.DateTime;
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpTransport;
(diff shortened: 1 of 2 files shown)


ewiseblatt:
@lwander

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -405,12 +413,19 @@ private void purgeOldVersions(String path) throws IOException {
...
   public long getLastModified(String daoTypeName) {
+      String path = daoRoot(daoTypeName) + '/' + LAST_MODIFIED_FILENAME;
       try {
-        return obj_api.get(
-                  bucketName,
-                  daoRoot(daoTypeName) + '/' + LAST_MODIFIED_FILENAME).execute()
-              .getUpdated().getValue();
+          return obj_api.get(bucketName, path).execute().getUpdated().getValue();
+      } catch (HttpResponseException e) {
+          if (e.getStatusCode() == 404) {
```
lwander: Won't this throw two 404's each time there is no last modified file?

ewiseblatt: Yes (though the second, which is in the update attempt to writeLastModified, appears to be a 400 not a 404).

However, unlike before (where there was only one 404 each time) this has the side effect of creating the file so that future calls will have 0 404's rather than 1 as before.

The net effect is a quick spike on startup with a new bucket but then going back to 0, whereas before there was a sustained failure rate until you create an artifact. My testing instance was sustained at 80% for the past few weeks because I'd clear things out after running  tests against it but leave the instance up. With this change, it blips to something like 4 errors (two for applications and two for pipelines) then goes to 0. Hardly noticeable and doesnt draw suspicion to investigate a systemic problem nor mask one that might be introduced thinking that a high error rate is normal to see.

lwander: Awesome, sounds good

lwander:
LGTM

PR #115, opened 2016-08-11
Handle parallel: "true" vs parallel: true

ajordens:


ajordens:
Ran into a few script-created pipelines that were using `parallel: "true"` rather than the expected `parallel: true`.

Migration will support either now.

PR #116, opened 2016-08-12
Support `lastModifiedBy` for GCS + S3

ajordens:


front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -237,6 +238,8 @@ public void deleteObject(String objectKey, String daoTypeName) {
 
   public <T extends Timestamped>
          void storeObject(String objectKey, String daoTypeName, T obj) {
+    obj.setLastModifiedBy(AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"));
```
ajordens: Didn't test this but it should work (analogous to S3)

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/ServiceAccountBucketDAO.java
```diff
@@ -41,7 +41,6 @@ public void update(String id, ServiceAccount permission) {
 
   private ServiceAccount upsert(String id, ServiceAccount permission) {
     permission.setName(id);
-    permission.setLastModified(System.currentTimeMillis());
```
ajordens: These are unnecessary calls as `lastModified` is determined by the object timestamp in GCS (and S3).

front50-web/src/main/groovy/com/netflix/spinnaker/front50/config/Front50WebConfig.groovy
```diff
@@ -45,6 +48,11 @@ public class Front50WebConfig extends WebMvcConfigurerAdapter {
   }
 
   @Bean
+  Filter authenticatedRequestFilter() {
+    new AuthenticatedRequestFilter(true)
```
ajordens: This is the filter that actually extracts the auth headers and sticks them in the MDC context.

duftler:
@ewiseblatt

ajordens:
FYI @anotherchrisberry 

@ewiseblatt I updated GCS (_it was an identical change to S3_) but didn't test it out.
@gregturn I _did not_ update the Redis implementation ... it felt boilerplate to do it in every single DAO. Was thinking it might be more elegant to have a serializer capable of setting `lastModifiedBy` (and `lastModified`) but figured you might have an even better spot.

cfieber:
LGTM

PR #120, opened 2016-08-18
Tactical application of hystrix to S3Support

ajordens:
- Avoids a scenario where S3 is unavailable for brief periods of time

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3Support.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3Support.java
@@ -21,6 +21,8 @@ import com.amazonaws.services.s3.model.*;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.front50.exception.NotFoundException;
+import com.netflix.spinnaker.front50.support.ClosureHelper;
+import com.netflix.spinnaker.hystrix.SimpleHystrixCommand;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/config/Front50WebConfig.groovy
```diff
@@ -71,4 +77,27 @@ public class Front50WebConfig extends WebMvcConfigurerAdapter {
...
     return new ItemDAOHealthIndicator(itemDAO: pipelineStrategyDAO)
   }
+
+  @Bean
+  HystrixRuntimeExceptionHandler hystrixRuntimeExceptionHandler() {
+    return new HystrixRuntimeExceptionHandler()
+  }
+
+  @ControllerAdvice
+  static class HystrixRuntimeExceptionHandler {
```
ajordens: Pretty close to what `igor`is doing

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3Support.java
```diff
@@ -103,20 +104,30 @@ public boolean isHealthy() {
...
-      T item = deserialize(s3Object);
-      item.setLastModified(s3Object.getObjectMetadata().getLastModified().getTime());
-      return item;
-    } catch (IOException e) {
-      throw new IllegalStateException(e);
-    } catch (AmazonS3Exception e) {
-      if (e.getStatusCode() == 404) {
-        throw new NotFoundException(String.format("No item found with id of %s", id.toLowerCase()));
-      }
+    return new SimpleHystrixCommand<T>(
```
ajordens: If there's an issue retrieving from S3, attempt to serve a previously cached value

ajordens:
Had a situation earlier this week where S3 was returning errors ... this PR introduces hystrix with a fallback for `findbyId` that will return a recently cached value in the event that the S3 lookup fails.

Hystrix exports a set of default metrics which makes it a little nicer than the try/catch alternative.

In this particular case, the failure to read from S3 resulted in a handful of failed deploys due to `clouddriver` not being able to retrieve the applications `legacyUdf` preference.

ajordens:
FYI @ewiseblatt 

PTAL @spinnaker/netflix-reviewers

tomaslin:
LGTM

PR #122, opened 2016-08-23
provider/core: Snapshot infrastructure saved as map rather than string

nwwebb:
I changed infrastructure to be saved as a map rather than a string. This makes it easier to parse the infrastructure into proper format for the diff view in deck. @lwander @danielpeach


lwander:
LGTM

lwander:
Please squash

lwander:
I let github do the squash

PR #126, opened 2016-08-30
Updates fiat api to disable csrf protection, fixes small SPeL error i…

ttomsu:
…n pipelines controller.

@duftler @skim1420 - this should fix your deleting pipeline/403 forbidden issues.


duftler:
Verified that I can now delete pipelines without getting the 403's.

PR #127, opened 2016-09-01
Optimizes calls to Fiat for getAll endpoints that can have large amou…

ttomsu:
…nts of returned data. Cleans up some batchUpdate TODOs.

Also removes a few `deleteById` endpoints because they're not being used as endpoints and would be difficult to authorize.

Metal Travis will choke until Fiat 0.5.0 is released.

@duftler @jtk54 @cfieber PTAL.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
@@ -43,6 +43,7 @@ class StrategyController {
     @Autowired
     PipelineStrategyDAO pipelineStrategyDAO
 
+    @PreAuthorize("@fiatPermissionEvaluator.storeWholePermission()")
     @PostFilter("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
     @RequestMapping(value = '', method = RequestMethod.GET)
     List<Pipeline> list() {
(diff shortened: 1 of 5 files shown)


ajordens:
It's not explicitly tied to this PR (_but I was away for the original_) ... but there is a Hystrix-related cache of all applications in `gate` that is used to serve fallbacks amongst other things. 

Not sure how this would play with request-level permissions applied to `list()` in `front50`?

jtk54:
@ajordens @ttomsu sounds like that may also need a Fiat annotation. Think we can annotate just the fallback function rather than duplicating the authZ check for both the Gate and Front50 endpoint.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
```diff
@@ -55,7 +56,7 @@ class StrategyController {
         pipelineStrategyDAO.getPipelinesByApplication(application)
     }
 
-    @PostAuthorize("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
+    @PostFilter("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
```
jtk54: Just for my understanding, does `PostFilter` run the function and then do the authZ check? Or the reverse?

jtk54:
LGTM. We should probably address the Hystrix caching business in Gate -- @ajordens is there any other Hystrix caching you're aware of?

ajordens:
There's also the business of `echo` pulling (_and caching_) the super set of all pipelines ... I think there was a plan around that but it escapes me.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/StrategyController.groovy
ttomsu: Yes - it evaluates the SpEL expression passed against the return object.

jtk54: Ok makes sense, we need that object's context before we can decide privileges.

ttomsu:
I've opened up two Fiat issues for Echo and Gate's caches. I'll tackle them in due time.

PR #132, opened 2016-09-14
Added Hystrix support

ewiseblatt:
@duftler 
This is only findById, but that's all that is on the S3 side.

@ajordens 
Are there other access points? Also, if we can do the refactoring we discussed over the summer then this should be easier to maintain (at least this case is in the system-independent part that is intended to be in the front50 core).

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/BucketDAO.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/BucketDAO.java
@@ -15,6 +15,8 @@
  */
 package com.netflix.spinnaker.front50.model;
 import com.netflix.spinnaker.front50.exception.NotFoundException;
+import com.netflix.spinnaker.front50.support.ClosureHelper;
+import com.netflix.spinnaker.hystrix.SimpleHystrixCommand;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/BucketDAO.java
```diff
@@ -107,8 +109,21 @@ public boolean isHealthy() {
 
 
   public T findById(String id) throws NotFoundException {
-      return service.loadCurrentObject(buildObjectKey(id), daoTypeName,
-                                       serializedClass);
+      return new SimpleHystrixCommand<T>(
+          getClass().getSimpleName(),
+          getClass().getSimpleName() + "-findById",
+          ClosureHelper.toClosure(args -> {
+              return service.loadCurrentObject(buildObjectKey(id), daoTypeName,
```
ajordens: Bit of an odd line break but otherwise LGTM.

```
return service.loadCurrentObject(
  buildObjectKey(id, daoTypeName, serializedClass)
);
```

^^ would be a little more consistent I think.

Same below with the String.format().

ajordens:
LGTM besides the aforementioned comment.

There are no other access points for now ... I agree on the refactoring. We're starting to make some API changes to `front50` (thanks to @jeyrschabu aka newest member of our team) so I suspect we'll actually be able to knock this off.

duftler:
A belated LGTM2.

PR #133, opened 2016-09-14
Changes getByName getter to return 404s if the app (NB: it's permissi…

ttomsu:
…on) is not found, as opposed to 403s when the app's permission doesn't exist.

This is necessary because Orca's UpsertApplicationTask first queries for the existence of an application (using the getByName() method) before invoking either the update() or create() endpoints on Front50.

tag: https://github.com/spinnaker/fiat/issues/51

@duftler @jtk54 PTAL


jtk54:
LGTM

ajordens:
LGTM

duftler:
LGTM2.

PR #135, opened 2016-09-15
Use `setUseSuffixPatternMatch` to fix dot truncation on parameters

ajordens:


ajordens:
FYI @jeyrschabu ... what was in there previously wasn't doing the trick.

PR #136, opened 2016-09-16
Fix an issue related to parameter names in V2 Application API

jeyrschabu:
- Reverted to using {applicationName:.+} to match url parameters containing a dot


jeyrschabu:
@ajordens I reverted to using the old scheme.

PR #138, opened 2016-09-19
Adds call to Fiat when application permission is edited to update use…

ttomsu:
…r permissions associated with that application.

Tests won't pass until 0.6.0 is cut and available, which requires https://github.com/spinnaker/fiat/pull/61 to be merged.

Closes https://github.com/spinnaker/fiat/issues/43


ttomsu:
Forgot to tag people to review :-/: @jtk54 @duftler

PR #140, opened 2016-09-22
deps update for ssl blacklist filehandle leak

cfieber:


ajordens:
We're already ahead of this one.

PR #144, opened 2016-09-28
Includes applications without explicit permission files when returnin…

ttomsu:
…g getAll

@jtk54 @duftler PTAL

@gbougeard FYI


jtk54:
LGTM

PR #149, opened 2016-10-13
Enable configuration values for Redis host and port

jrefi:
The Redis backend defaulted to localhost:6379. This PR enables specifying an arbitrary Redis host/port.

### DIFF
--- a/front50-redis/src/main/groovy/com/netflix/spinnaker/front50/redis/RedisConfig.groovy
+++ b/front50-redis/src/main/groovy/com/netflix/spinnaker/front50/redis/RedisConfig.groovy
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.front50.model.project.Project
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.data.redis.connection.RedisConnectionFactory
 import org.springframework.data.redis.connection.jedis.JedisConnectionFactory
 import org.springframework.data.redis.core.RedisTemplate
@@ -31,6 +32,12 @@ import org.springframework.data.redis.serializer.StringRedisSerializer
 @Configuration
 @ConditionalOnExpression('${spinnaker.redis.enabled:false}')
 class RedisConfig {
+
+  @Value('${spinnaker.redis.host:localhost}')
+  private String host;
+
+  @Value('${spinnaker.redis.port:6379}')
+  private Integer port;
 
   @Bean
   RedisApplicationDAO redisApplicationDAO(RedisTemplate<String, Application> template) {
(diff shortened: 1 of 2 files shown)


duftler:
@gregturn PTAL

gregturn:
Given that front50's redis is using Spring Data Redis + Spring Boot, you can simply override things through `RedisProperties`:

```
# REDIS (RedisProperties)
spring.redis.database= # database name
spring.redis.host=localhost # server host
spring.redis.password= # server password
spring.redis.port=6379 # connection port
spring.redis.pool.max-idle=8 # pool settings ...
spring.redis.pool.min-idle=0
spring.redis.pool.max-active=8
spring.redis.pool.max-wait=-1
spring.redis.sentinel.master= # name of Redis server
spring.redis.sentinel.nodes= # comma-separated list of host:port pairs
```

Not sure we need this. Given this is Spring Boot, probably simply need to remove the RedisConnectionFactory in RedisConfig, and instead use Spring Boot's autoconfigured one from `RedisAutoConfiguration`

jrefi:
@gregturn I am deploying Spinnaker on top of k8s. How would I go about change the RedisProperties in that case?

gregturn:
@jrefi I believe the fault lies in another piece of spinnaker that actually uses `redis.connection` as the @Value for connecting to Spinnaker. And on PCF, the Java Buildpack automatically swaps out the connection factory when you bind to a redis instance.

Essentially, you can either drop such property settings in front50-local.yml or set env variables to override properly.

You may wish to read http://docs.spring.io/spring-boot/docs/1.2.8.RELEASE/reference/htmlsingle/#boot-features-external-config to see full gamut of options when overriding property settings with Spring Boot.

jrefi:
> But that presumes we use Spring Boot's autoconfigured JedisConnectionPool and JedisConnectionFactory. Looks like front50 is supplying a manually created one which will cause Spring Boot to back off using the autoconfigured one.

So there is a bug in the Redis driver for front50? Until that is remedied can my PR still be accepted as this is blocking us from being able to deploy Spinnaker internally.

gregturn:
My proposal is to not use your code change but instead removing the beans I listed. Then you'll have standard properties to override.

jrefi:
I searched around the project and found other examples where the Redis connection is created in a similar manner.

From GateConfig

`  @Bean
  JedisConnectionFactory jedisConnectionFactory(
    @Value('${redis.connection:redis://localhost:6379}') String connection
  ) {
    URI redis = URI.create(connection)
    def factory = new JedisConnectionFactory()
    factory.hostName = redis.host
    factory.port = redis.port
    if (redis.userInfo) {
      factory.password = redis.userInfo.split(":", 2)[1]
    }
    factory
  }`

How is what I am doing different from that? I can change my code to read from the property ${redis.connection} if that is preferred.

gregturn:
Front50 is the only one using spring data Redis. If it was up to me, I'd update all modules to use the same approach including migrating to spring data Redis. 

I'm not a committer so it's not up to me. This is my opinion for what's best on front50.

ajordens:
TBH I don't see a strong reason to not support this in `front50`.

PR #151, opened 2016-10-25
First stab at storing tagged entities in Front50

ajordens:
- S3 only for now

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/EntityTags.groovy
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/EntityTags.groovy
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.tag
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter
+import com.fasterxml.jackson.annotation.JsonAnySetter
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.netflix.spinnaker.front50.model.Timestamped
+
+class EntityTags implements Timestamped {
+  String id
+  String idPattern   // pattern used to construct the `id`, ie. {{entityType}}__{{entityId}}__{{account}}__{{region}}
+
+  Long lastModified
+  String lastModifiedBy
+
+  Map<String, Object> tags = [:]
+  EntityRef entityRef
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  static class EntityRef {
+    private Map<String, Object> attributes = new HashMap<String, Object>()
+
+    String entityType
+    String entityId
+
+    @JsonAnyGetter
+     Map<String,Object> attributes() {
+      return attributes;
+    }
+
+    @JsonAnySetter
+    void set(String name, Object value) {
+      attributes.put(name, value);
+    }
+  }
+}
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
@@ -126,4 +126,9 @@ public class S3Config {
   public S3SnapshotDAO s3SnapshotDAO(ObjectMapper objectMapper, AmazonS3 amazonS3) {
     return new S3SnapshotDAO(objectMapper, amazonS3, Schedulers.from(Executors.newFixedThreadPool(5)), 10000, bucket, rootFolder);
   }
+
+  @Bean
+  public S3EntityTagsDAO s3EntityTagsDAO(ObjectMapper objectMapper, AmazonS3 amazonS3) {
+    return new S3EntityTagsDAO(objectMapper, amazonS3, null, -1, bucket, rootFolder);
+  }
 }
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/exception/BadRequestException.groovy
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/exception/BadRequestException.groovy
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.exception
+
+import groovy.transform.InheritConstructors
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ResponseStatus
+
+@InheritConstructors
+@ResponseStatus(value = HttpStatus.BAD_REQUEST)
+class BadRequestException extends RuntimeException {}
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/EntityTagsDAO.groovy
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/EntityTagsDAO.groovy
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.tag
+
+import com.netflix.spinnaker.front50.model.ItemDAO
+
+interface EntityTagsDAO extends ItemDAO<EntityTags> {
+  /**
+   * @param prefix TaggedEntity.id prefix (application/front50, etc.)
+   * @return All tagged entities with an id beginning with {@code prefix}
+   */
+  Set<EntityTags> all(String prefix)
+}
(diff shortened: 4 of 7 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -126,4 +126,9 @@ public S3PipelineDAO s3PipelineDAO(ObjectMapper objectMapper, AmazonS3 amazonS3)
   public S3SnapshotDAO s3SnapshotDAO(ObjectMapper objectMapper, AmazonS3 amazonS3) {
     return new S3SnapshotDAO(objectMapper, amazonS3, Schedulers.from(Executors.newFixedThreadPool(5)), 10000, bucket, rootFolder);
   }
+
+  @Bean
+  public S3TaggedEntityDAO s3TaggedEntityDAO(ObjectMapper objectMapper, AmazonS3 amazonS3) {
+    return new S3TaggedEntityDAO(objectMapper, amazonS3, null, -1, bucket, rootFolder);
```
ajordens: Differs slightly from other DAOs in that it will _not_ cache all entities in-memory (_thus no scheduler and -1 refresh interval_)

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3TaggedEntityDAO.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Override
+  public Set<TaggedEntity> all(String prefix) {
+    ObjectListing bucketListing = amazonS3.listObjects(
+      new ListObjectsRequest(bucket, (rootFolder + prefix).toLowerCase(), null, null, 100)
+    );
+    List<S3ObjectSummary> summaries = bucketListing.getObjectSummaries();
+
+    // TODO-AJ this is naive and inefficient
```
ajordens: Naive but there's a max of 100 objects ... plan on revisiting this down the road (_unsure how common this endpoint would actually be called_)

ajordens:
PTAL @spinnaker/netflix-reviewers (FYI @spinnaker/google-reviewers)

This is an early effort at supporting _generic_ tagging in Spinnaker, the current thinking is that we will use `front50` as a durable store and `elasticsearch` as an efficient search index.

Tagging functionality is expected to be fronted by `clouddriver`:
1) `clouddriver` writes to `front50`
2) `clouddriver` writes to `elasticsearch`
3) `clouddriver` reads only from `elasticsearch` or a mix of `elasticsearch` and `front50` depending on the particular use case (_needs additional testing_)

@spinnaker/google-reviewers - Happy to share a document that outlines our various use cases for generic tagging ... in particular we're looking to tag things that are otherwise unsupported by `aws` (_IAMRoles, clusters, etc._).

Once things have been proven out a bit, I'll add support for `gcs` ... possibly doing the migration/refactor to use the Bucket API.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/TaggedEntity.groovy
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+
+package com.netflix.spinnaker.front50.model.tag
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.netflix.spinnaker.front50.model.Timestamped
+
+class TaggedEntity implements Timestamped {
+  String id
```
ajordens: id is of the form `{entityType}/{entityId}/{account}/{region}`:

- `servergroup/front50-main-v001/prod/us-west-1`
- `cluster/front50-main/prod/*`
- `application/front50/*/*`

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/TaggedEntityDAO.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.front50.model.tag
+
+import com.netflix.spinnaker.front50.model.ItemDAO
+
+interface TaggedEntityDAO extends ItemDAO<TaggedEntity> {
+  Set<TaggedEntity> all(String prefix)
```
ajordens: `prefix` is something along the lines of:

- `application/front50`
- `cluster/front50-main`

It will match any tagged entity with a id _starting with_ `prefix`

ewiseblatt:
Could we refactor S3 and GCS so they share most of the implementation? I dont recall the details anymore, but though I had tee'd that up. This is going to diverge and make the refactoring harder while also diverging functionality.

If you cant attempt it, I can take a stab at it tomorrow if you can hold off on this to see how that turns out. Or we can talk about it first if you want. I can even talk now if you want though am not prepared with off the cuff details. I'm available any time tomorrow except 11:15-12:15 EST.

ajordens:
@ewiseblatt I was going to do it after I get this proved out ... looked a bit this afternoon to refresh my memory and I suspect it will still be reasonably straight forward. 

At a minimum I'll need to refactor/move a few of the common interfaces out of `front50-gcs`.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/TaggedEntity.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  static class EntityRef {
+    String entityType
+    String entityId
+
+    String cloudProvider
+    String application
+    String account
+    String region
```
ewiseblatt: what is an entity?
would gcs need a zone for it?

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3TaggedEntityDAO.java
ewiseblatt: You should be able to indicate that there were more than 100.
Even if that means listing 101 and seeing if there is a 101st and only returning the first 100.

ajordens: That's reasonable ... I anticipate this endpoint (_at least as it is right now_) is purely for debugging.

ewiseblatt: what is a tagged entity? sounds familiar.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/TaggedEntity.groovy
ajordens: An `EntityRef` is an encapsulation of the coordinates needed to lookup tagged object.

A tagged object could be anything:
- cluster
- server group
- iam role
- load balancer
- security group

Each of these would have slightly different coordinates and at a minimum some combination of:
- account
- region
- cloudprovider

I'm not sure the granularity of things you would want to tag in GCP ... @duftler? 

I'm considering changing `EntityRef` to use `@JsonAnyGetter`/`@JsonAnySetter` and removing the explicit attributes.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3TaggedEntityDAO.java
ajordens: An entity (server group/cluster/iamrole/security group) w/ tags.

The entity itself isn't stored, only coordinates that could be used to look it up in `clouddriver`.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/TaggedEntity.groovy
duftler: The list of tagged objects you gave as an example seems like the level of granularity I would expect.

And I like the idea of not having a fixed set of attributes in `EntityRef`.

duftler:
Would be very interested in seeing the use-cases doc you mentioned.

duftler:
Also curious about how this spinnaker-level tagging will intersect with whatever tagging/labeling is supported at the individual platform levels.

ajordens:
@duftler Shared a WIP doc that outlines (_some of_) the use cases that we're aware of

I'm not sure they'll be an effort (_frankly not sure it's sensible or feasible_) to push Spinnaker-level tags down to the individual platform-level. There are limitations (_number of tags and size of individual tags_) that make this difficult, not to mention that individual platforms don't universally support all that is desired to be tagged.

I anticipate there being situations where tagging server groups at the platform-level is desirable (integration with platform features or other systems that speak platform tags), and for that I'd expect the existing platform-specific operations to be used ... much like they are right now.

I think it would make sense to automatically fold Spinnaker-level tags into `individual` results ... some additional testing needs to be done to determine what makes sense for folding tags into lists of results (_ie. search for all server groups_).

Happy to hear feedback or have a discussion around ^^.

ajordens:
`EntityRef` now uses `@JsonAnyGetter/Setter` to track arbitrary attributes.

ewiseblatt:
@duftler, for that matter, would cluster be relevant?

ewiseblatt:
Yes, we'll need to move those out of front50-gcs at a minimum. It may be
that there are additional cleanups that can happen as well but at the time,
that's all it seemed to be. The modules have evolved since that point in
time so I'm not sure where things actually are, which is all the more
reason to bring them together.

I think it will be easier to do the refactoring first. Not just because
this will diverge, but my gut tells me that if you map from GCS back into
S3 then compare what you currently have against that then that would be the
most direct route. And from a self interest standpoint, it would be much
simpler on me trying to keep up. I'd rather invest my effort in the
refactoring to bring us together than in duplicating effort to maintain
parity.

I havent looked at the details of your PR yet. I'll probably do that in the
morning but I might come back to it after I take a break.

- eric

ewiseblatt:
What do you tag and where do you tag it? Why (I can imagine things I'd do
with this, but asking for the official motivation to help guide my
thinking).
How are tags managed with respect to the referenced entity?
If I tag a server group then destroy the server group, what happens to the
tag?
Though you call it a "TaggedEntity" it looks like "EntityTags" where the
entity doesnt know about the tags, but the tags are managed orthogonal to
it. If so, then you might want some kind of administrative ability to find
instances with invalid references. The underlying entity's lifecycle is
independent.

ajordens:
`EntityTags` makes more sense than `TaggedEntity`, I like that.

Current thoughts are that the tagging is something orchestrated through a clouddriver/orca operation.

There is certainly a lifecycle around particular entities ... I would hope that anything deleted via Spinnaker would be able to also delete tags. There are also timestamps that could be used to filter tags. 

Lastly, we might need a periodic reconciler that'll remove any tags whose associated entity no longer exists. This would be a common case from something like `janitor monkey`.

ajordens:
One concrete use case we're looking at is programmatically tagging IAMRoles to restrict which applications can launch with them.

cfieber:
LGTM

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/EntityTags.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  static class EntityRef {
+    private Map<String, Object> attributes = new HashMap<String, Object>()
+
+    String entityType
+    String entityId
+
+    @JsonAnyGetter
+     Map<String,Object> attributes() {
```
ewiseblatt: Before the attributes were just strings. Now it looks like they can be composite. Does that pose a problem querying them back? Just asking. I didnt look up what JsonAnyGetter is.

ajordens:
@ewiseblatt I'm going to start on the refactor now

ewiseblatt:
I took too long in my pass and you already merged. That's ok but my questions still hold. If you want to discuss refactoring while you are doing that I'm game for that as well.

ajordens:
The entity id itself is deterministically generated by `clouddriver` (likely using the existing `Keys` classes), so fetching tasks for a single entity is straight forward.

Delete by `id` should work similarly. 

The `@JsonAnyGetter/Setter` change above just allows us to include any arbitrary key/value pair in an entity ref (_vs. the explicit account/region/cloudProvider that was in there previously_). The change from `String` -> `Object` was to support the notion that not all tag values need to be a `String` (numeric and boolean seem reasonable).

PR #152, opened 2016-10-26
Merging `gcs` and `s3` access patterns

ajordens:
The idea being that we can have a series of default DAOs that are backed by an s3 or gcs `storageService`

- Renamed `BucketDAO` to `StorageServiceSupport`
- Encapsulated `serializedClass` and `daoTypeName` in `ObjectType`


front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -94,66 +94,58 @@ public RestTemplate restTemplate() {
 
   @Bean
   public ApplicationBucketDAO applicationDAO(GcsStorageService service) {
-    return new ApplicationBucketDAO(rootFolder,
-                                    service,
+    return new ApplicationBucketDAO(service,
```
ajordens: `rootFolder` was never used/needed

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/ApplicationBucketDAO.java
```diff
@@ -25,13 +25,11 @@
 import java.util.Collection;
 import java.util.Map;
 
-public class ApplicationBucketDAO extends BucketDAO<Application> implements ApplicationDAO {
-   public ApplicationBucketDAO(String basePath,
-                               StorageService service,
+public class ApplicationBucketDAO extends StorageServiceSupport<Application> implements ApplicationDAO {
```
ajordens: These DAOs will no longer exist after this refactor (rather we'll have *one* of each type backed by a `StorageService`.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -187,6 +187,7 @@ public void ensureBucketExists() {
   /**
    * Returns true if the storage service supports versioning.
    */
+  @Override
```
ajordens: I don't believe I changed the functionality of anything in here but I'll need you to give it a whirl once things are closer to landing (probably tomorrow sometime).

This differs slightly from `s3` due to the difference in naming around filenames (`specification.json` vs `application-metadata.json` and `last-modified.json` vs. `last-modified`).

Also added missing @Override's

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
```diff
@@ -0,0 +1,221 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class S3StorageService implements StorageService {
```
ajordens: Tried to keep this as similar to what exists in `S3Support`

ajordens:
WIP -- do not merge (tests don't pass anyways).

FYI @ewiseblatt ... this is where I got to after some effort merging what has been individually done across `gcs` and `s3` implementations.

I've verified that we can still:
- fetch all applications from s3
- fetch a single application from s3
- delete an application from s3

I still need to create default DAOs for each of the types and test each one individually.

Once that's all done, I plan on deleting the `s3`-specific code.

ewiseblatt:
The gist looks good so far. I didnt look at the S3 details yet. I checked out your branch, deployed it, and ran my GCS integration tests against it and they worked. I actually wasnt expecting them to because it looked like you changed the file names written out and my test peeks at the actual files written. I wasnt expecting it to find them. I guess that is encapsulated into the GcsStorageService part which isnt affected.

ajordens:
Nice, thanks for the quick sanity check @ewiseblatt. The filenames are encapsulated within `GcsStorageService` so I wouldn't expect the behavior to change ... but I'm glad you were able to validate that.

Have a great night!

ewiseblatt:
FYI I just tried another sync and integration test. I only have application and pipelines because the implementation and mechanisms were the same so it didnt seem necessarily to test them all. So given those two, the pipeline tests all pass, but the application tests all fail because the POST methods are returning 405 errors.

2016-10-26 17:38:58.926	collapsePOST http://localhost:53904/default/applications/name/gcpfront50test173847
{
  "updateTs": "1463667655844",
  "cloudProviders": "gce,aws",
  "platformHealthOnly": false,
  "description": "My Application Description.",
  "createTs": "1463666817476",
  "accounts": "my-aws-account,my-google-account",
  "email": "test@google.com",
  "name": "gcpfront50test173847"
}
2016-10-26 17:38:59.046	collapseHTTP 405
{
  "status": 405,
  "timestamp": 1477517939020,
  "message": "Request method 'POST' not supported",
  "error": "Method Not Allowed"
}

2016-10-26 17:39:04.682	collapsePUT http://localhost:53904/default/applications
{
  "updateTs": "1463667655845",
  "name": "gcpfront50test173847",
  "createTs": "1463666817477",
  "accounts": "NEW_my-aws-account,my-google-account",
  "cloudProviders": "gce,aws,kubernetes",
  "email": "NEW_test@google.com",
  "description": "NEW_My Application Description."
}
2016-10-26 17:39:04.799	collapseHTTP 405
{
  "status": 405,
  "timestamp": 1477517944778,
  "message": "Request method 'PUT' not supported",
  "error": "Method Not Allowed"
}

2016-10-26 17:39:19.451	collapseDELETE http://localhost:53904/default/applications/name/gcpfront50test173847
2016-10-26 17:39:19.549	collapseHTTP 405
{
  "status": 405,
  "timestamp": 1477517959529,
  "message": "Request method 'DELETE' not supported",
  "error": "Method Not Allowed"
}


Where as this works:
2016-10-26 17:39:15.331	collapsePOST http://localhost:53904/pipelines
{
  "index": 0,
  "name": "My gcpfront50test173847 Pipeline",
  "triggers": [],
  "limitConcurrent": true,
  "keepWaitingPipelines": false,
  "parallel": true,
  "lastModifiedBy": "anonymous",
  "application": "gcpfront50test173847",
  "stages": [],
  "id": "gcpfront50test173847-pipeline-id"
}
2016-10-26 17:39:16.743	HTTP 200

ajordens:
Can you update your application tests to hit `/v2/applications/` ... I removed the old `/default/applications` endpoint as part of this commit as it had been deprecated a month or so back (nothing in gate/clouddriver is hitting it anymore).

ajordens:
POST /v2/applications

ajordens:
I've tested the S3 bits locally alongside `gate` and `orca`, it all looks good.

I'm ready to merge when you're satisfied @ewiseblatt

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -197,19 +198,25 @@ public boolean supportsVersioning() {
     }
   }
 
+  @Override
+  public <T extends Timestamped> Collection<T> loadObjectsWithPrefix(ObjectType objectType, String prefix, int maxResults) {
+    throw new UnsupportedOperationException("loadObjectsWithPrefix is not yet supported!");
```
ajordens: This will need a GCS implementation @ewiseblatt ... only used by `EntityTags` so not breaking any existing functionality.

ewiseblatt:
The thing I grabbed earlier works with my tests.
I'm going to review the actual PR.
I'll try to review this by late morning.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ObjectType.java
```diff
@@ -0,0 +1,47 @@
...
+import com.netflix.spinnaker.front50.model.serviceaccount.ServiceAccount;
+import com.netflix.spinnaker.front50.model.snapshot.Snapshot;
+import com.netflix.spinnaker.front50.model.tag.EntityTags;
+
+public enum ObjectType {
+  PROJECT(Project.class, "projects", "project-metadata.json"),
+  PIPELINE(Pipeline.class, "pipelines", "pipeline-metadata.json"),
+  STRATEGY(Pipeline.class, "pipeline-strategies", "pipeline-strategy-metadata.json"),
+  NOTIFICATION(Notification.class, "notifications", "notification-metadata.json"),
+  SERVICE_ACCOUNT(ServiceAccount.class, "serviceAccounts", "serviceAccount-metadata.json"),
```
ajordens: `SERVICE_ACCOUNT`, `APPLICATION_PERMISSION` and `SNAPSHOT` do not fit the typical naming conventions for metadata.

I have no idea if anyone on aws is using these otherwise I'd be inclined to make them consistently named (metadata filename in particular) ... nobody at Netflix is using them.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -34,58 +33,49 @@
...
-            refresh();
-          } catch (Exception e) {
-            log.error("Unable to refresh: {}", e);
-          }
-        });
-    Observable
+    if (refreshIntervalMs > 0) {
+      try {
+        log.info("Warming Cache");
+        refresh();
```
ajordens: I did swap this such that it runs on the main thread ... I don't believe it'll materially increase boot time.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/DefaultPipelineStrategyDAO.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public Collection<Pipeline> getPipelinesByApplication(String application) {
+    return all()
+      .stream()
+      .filter(pipelineStrategy -> pipelineStrategy.getApplication().equalsIgnoreCase(application))
+      .collect(Collectors.toList());
+  }
+
+  @Override
+  public Collection<Pipeline> getPipelineHistory(String id, int limit) {
```
ewiseblatt: here you call it getPIpelineHistory
in snapshot below you are just getHistory
in notification there is no history.

When is there and is there not history?
At least in GCS, history is a bucket level so in fact everythig nwill have a history whether or not we access it.
Why not just add getHistory to the base StorageServiceSupport? These methods just pass through anyway. You can override when you cannot support history for whatever reason, or need to do so differently.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/DefaultEntityTagsDAO.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private EntityTags upsert(String id, EntityTags tag) {
+    Objects.requireNonNull(id);
+    super.update(id, tag);
+    return findById(id);
+  }
+
+  @Override
+  public Collection<EntityTags> all() {
+    // no support for retrieving _all_ tagged entities
```
ewiseblatt: why is that?
Is an unsupported error more useful?
An empty 200 sounds like it will lead to confusion.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/project/DefaultProjectDAO.java
```diff
@@ -0,0 +1,56 @@
...
+      id = UUID.randomUUID().toString();
+    }
+    item.setId(id);
+
+    update(id, item);
+    return findById(id);
+  }
+
+  @Override
+  public void truncate() {}
```
ewiseblatt: what is this?

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -83,52 +102,55 @@ public RestTemplate restTemplate() {
...
+    AmazonObjectMapperConfigurer.configure(awsObjectMapper);
+
+    return new S3StorageService(awsObjectMapper, amazonS3, bucket, rootFolder);
   }
 
   @Bean
-  public S3ApplicationDAO s3ApplicationDAO(ObjectMapper objectMapper, AmazonS3 amazonS3) {
-    return new S3ApplicationDAO(objectMapper, amazonS3, Schedulers.from(Executors.newFixedThreadPool(20)), 15000, bucket, rootFolder);
+  public ApplicationDAO applicationDAO(StorageService storageService) {
+    return new DefaultApplicationDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(20)), 15000);
```
ewiseblatt: How are the rates determined? Maybe these should be factored out as well. And possibly the number of threads. I imagine these are more behavioral (nature of the type and how it is used in spinnaker) than platform (s3 vs gcs). Maybe there can be a "small" and "large" config where one is for small projects that dont change frequently and one is for large projects that have a lot of data and changes more often. If that affects the tradeoffs made here.

Maybe not now, but ultimately it seems these might need to be tweakable for the extreme cases. Unless perhaps the rate is adaptable to the size of the domain or the update time (i.e. not constant). But that's not for this PR and depends on whether a suitable default for the extremes in use cases exists.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
```diff
@@ -0,0 +1,253 @@
...
+  @Override
+  public <T extends Timestamped> T loadObjectVersion(ObjectType objectType,
+                                                     String objectKey,
+                                                     String versionId) throws NotFoundException {
+    throw new UnsupportedOperationException("loadObjectVersion() is not supported");
+  }
+
+  @Override
+  public void deleteObject(ObjectType objectType, String objectKey) {
+    amazonS3.deleteObject(bucket, buildS3Key(objectType.group, objectKey, objectType.defaultMetadataFilename));
```
ewiseblatt: In create you wrap errors but not here.
I chose to ignore 404. We should be consistent, so I assume you'd want the same. If not I'll take mine out. Unless you wish you could 404 but cannot for legacy reasons.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/EntityTagsController.groovy
```diff
@@ -33,17 +36,18 @@ import org.springframework.web.bind.annotation.RestController
...
 @RequestMapping(value = "/v2/tags", produces = MediaType.APPLICATION_JSON_VALUE)
-class TagController {
+class EntityTagsController {
   @Autowired(required = false)
   EntityTagsDAO taggedEntityDAO
 
   @RequestMapping(method = RequestMethod.GET)
   Set<EntityTags> tags(@RequestParam(value = "prefix", required = true) String prefix) {
-    return taggedEntityDAO?.all(prefix) ?: []
+    return taggedEntityDAO?.all(prefix, 100) ?: []
```
ewiseblatt: I thought taggedEntityAll returned empty.

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/NotificationControllerTck.groovy
```diff
@@ -188,15 +189,17 @@ class S3NotificationControllerTck extends NotificationControllerTck {
...
 
   @Override
   NotificationDAO createNotificationDAO() {
     def amazonS3 = new AmazonS3Client(new ClientConfiguration())
     amazonS3.setEndpoint("http://127.0.0.1:9999")
     S3TestHelper.setupBucket(amazonS3, "front50")
 
-    s3NotificationDAO = new S3NotificationDAO(new ObjectMapper(), amazonS3, scheduler, 0, "front50", "test")
-    return s3NotificationDAO
+    def storageService = new S3StorageService(new ObjectMapper(), amazonS3, "front50", "test")
```
ewiseblatt: I'm not sure what is going on here.
Is it possible/practical to test S3Storage service separate and write these tests in terms of the base controller?

Ideally this is base controller and there is another that tests StorageService where we can each plug our StorageService in and it verifies compliance (throw consistent exceptions, returns consistent results). I dont know how hard that is. i dont want to compromise your testing, but would like mine to benefit and ensure we stay in sync.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/DefaultPipelineStrategyDAO.java
ajordens: Chalk it up to different implementations by the Netflix (pipeline/application, etc.) and Google (snapshot, applicationpermission, etc.).

I didn't go out of my way as part of this PR to fundamentally change the interface methods ... I don't see a reason why it shouldn't  be `getHistory()` (still would say do it as a separate PR)

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/project/DefaultProjectDAO.java
ajordens: We have some internal endpoints that let you do delete everything in the bucket (for example). 

Not exposed outside of `front50` ... not sure even used.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/tag/DefaultEntityTagsDAO.java
ajordens: Not sure ... tags are a bit of a special case and there is no actual caller of this DAO method.

The controller itself doesn't have an equivalent all() endpoint (it requires a query parameter).

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/DefaultPipelineStrategyDAO.java
ewiseblatt: ok. This is easily deprecated by having the old method delegate to the new then taking it out at some point down the road. If history were always available in the base class then it would simplify how you think about and test it. But you can do it in a separate PR. Just mentioning it here while we're discussing it.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/EntityTagsController.groovy
ajordens: This is an overloaded `all` with a parameter ...

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/NotificationControllerTck.groovy
ajordens: These were some previous tests that could be pointed at an S3 proxy running locally on a laptop.

Honestly rarely ran beyond when S3 support was originally added.

ewiseblatt: I think the right thing to do here is to write the tests in terms of the StorageService notification so that you are testing the model is correct for the data model / resource implementations themselves.

Then we need to write a unit test suite to test the specialized StorageService implementations. As I mentioned before, ideally there is a base suite that we inject a concrete class into and it verifies the contract. Each service can add additional tests beyond that for the extra/special behaviors on it. That's what I've done in the past for similar situations and it turns out pretty easy. Except here there is the extra underlying service calls that would need to be mocked with spock so there's a twist that might not be so straightforward to reuse a standard test suite.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/DefaultPipelineStrategyDAO.java
ajordens: Just pushed another commit that pulled `history(id, maxResults)` up to `ItemDAO` and out of the individual DAOs ... it should be a little more consistent.

Not convinced that this won't change again in the future but it'll be an easier job next time (less code).

ewiseblatt:
I cant figure out this github interface.
I want to say that ApplicationsController is still calling getApplicationHistory but that method no longer exists. It looks like it is related to that last push you did.

ewiseblatt:
The same appears true for getPipelineHistory.

ewiseblatt:
Is pipelines still v1 or should I be calling a v2 interface there? I'm calling v2 on applications and v1 on pipelines. Since pipelines broke just now, I assume I'm calling the right interface.

ajordens:
Pipelines still lives under / ... although it's api should be reviewed and moved to `/v2` when it's access patterns can be made reasonably consistent with the others.

GET `/v2/pipelines`
GET `/v2/pipelines?application={applicationName}`
GET `/v2/pipelines/{pipelineId}` 
GET `/v2/pipelines/{pipelineId}/history` 
PUT `/v2/pipelines/{pipelineId}` 
DELETE `/v2/pipelines/{pipelineId}` 
... etc.

`pipelineId` should be an actual UUID and not a pipeline name ... I believe we're somewhat inconsistent around this.

ewiseblatt:
LGTM

As I noted offline, it would be nice to make the unit tests of the core components use the base StorageService interface to verify the protocol and such. Then separately have a reusable test fixture that can test the basic protocol against each of the specialized StorageServices. The implementations would then inject some testing thing into that test plus whatever extensions they have.

Now is the best time to introduce that.

PR #153, opened 2016-10-31
Adds 'runAsUser' service account permission checking as prerequisite …

ttomsu:
…to saving a pipeline.

Tests won't pass until https://github.com/spinnaker/fiat/pull/108 is merged and released.

tag: https://github.com/spinnaker/fiat/issues/36

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/AuthorizationSupport.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/AuthorizationSupport.groovy
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.front50.controllers
+
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator
+import com.netflix.spinnaker.front50.model.pipeline.Pipeline
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.security.core.Authentication
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.stereotype.Component
+
+@Component
+class AuthorizationSupport {
+
+  @Autowired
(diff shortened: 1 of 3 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/AuthorizationSupport.groovy
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class AuthorizationSupport {
+
+  @Autowired
+  FiatPermissionEvaluator permissionEvaluator
+
+  boolean hasRunAsUserPermission(Pipeline pipeline) {
+    List<String> runAsUsers = pipeline.triggers*.runAsUser
```
jtk54: Not all pipelines have `triggers`, right? Might need a `?` and/or a `?: []`.

ttomsu: My testing of the spread operator shows it to be null safe.

https://groovy-playground.appspot.com/#?load=8493a5b4e9ebea349d4d99705e964ef2

ajordens:
So you're basically enforcing that if we're attempting to save a pipeline w/ a `runAsUser` trigger that the authenticated user actually has access to the trigger user?

Do strategies need similar treatment?

ajordens:
Sorry, didn't mean to close.

ttomsu:
Your understanding is correct. We have to do this on pipeline save so that Echo doesn't have to do (and handle the failure of) an auth check.

A cursory look at strategy config doesn't show an ability to trigger a strategy outside of a "deploy" stage. So no, I don't think you'd be able to alter the trigger user from a strategy.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -46,10 +46,10 @@ class PipelineController {
   @Autowired
   PipelineDAO pipelineDAO
 
-  @PreAuthorize("@fiatPermissionEvaluator.storeWholePermission()")
-  @PostFilter("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
+  @PreAuthorize("#restricted ? @fiatPermissionEvaluator.storeWholePermission() : true")
+  @PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
   @RequestMapping(value = '', method = RequestMethod.GET)
-  List<Pipeline> list() {
+  List<Pipeline> list(@RequestParam(required = false, value = 'restricted', defaultValue = 'true') boolean restricted) {
```
ajordens: Thoughts on this as an HTTP header vs query parameter? Former might align better given that everything else auth-related flows over headers?

ttomsu: Either one would work. I opted for a query param for two reasons:

1.) the ?restricted=false appears in the logs by default
2.) consistent with the same thing listing Applications in Clouddriver.

I plan to submit as is, and can file a bug to move both servers and clients over to headers if you want.

PR #158, opened 2016-11-07
Calls Fiat sync when service accounts are created or deleted.

ttomsu:
@duftler @jtk54 @ajordens @cfieber PTAL

Fixes https://github.com/spinnaker/fiat/issues/111

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.groovy
@@ -16,8 +16,11 @@
 
 package com.netflix.spinnaker.front50.controllers
 
+import com.netflix.spinnaker.fiat.shared.FiatService
 import com.netflix.spinnaker.front50.model.serviceaccount.ServiceAccount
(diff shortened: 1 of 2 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.groovy
```diff
@@ -34,18 +39,39 @@ public class ServiceAccountsController {
...
     serviceAccountDAO.delete(serviceAccountId)
+    syncUsers(acct)
+  }
+
+  private void syncUsers(ServiceAccount serviceAccount) {
+    if (!fiatService || !serviceAccount) {
+      return
+    }
+
+    def roles = [StringUtils.substringBefore(serviceAccount.name.replaceFirst('%40', "@"), "@")]
```
jtk54: Is `serviceAccount.name` always url-encoded? `replaceFirst` returns the string whether it replaces a character or not -- might be an issue here.

ttomsu: It's not always encoded. Can you think of a common use case where this would be an issue?

PR #160, opened 2016-11-11
Adding s3 bucket failover configuration and readonly mode

robzienert:
Adds configuration for a failover bucket and region for S3 to support internal availability stories. When `spinnaker.s3.failover.enabled = true`, the application will connect to the configured failover bucket, as well as set front50 into a read-only mode. 

For the foreseeable future, this flag will be manually managed and require service restart to flip.


cfieber:
LGTM

PR #162, opened 2016-11-28
Adding support for searching project by a variety of keywords

jeyrschabu:
- Added support for a fuzzy string match on object attributes
- Added support for pageSize similar to applications
- Search is case insensitive on key and value and allows searching even deep attributes on the object
- Using the same ranking algo for search result relevance as applications (Worth to note that this doesnt apply to deep properties)

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.front50.controllers.v2
 
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.front50.exception.NotFoundException
+import com.netflix.spinnaker.front50.model.SearchUtils
 import com.netflix.spinnaker.front50.model.project.Project
 import com.netflix.spinnaker.front50.model.project.ProjectDAO
 import groovy.transform.InheritConstructors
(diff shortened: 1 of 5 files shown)


jeyrschabu:
@ajordens PTAL

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.groovy
```diff
@@ -72,10 +73,14 @@ public class ProjectsController {
     return projects
   }
 
-  @ApiOperation(value = "", notes = "Fetch all projects")
+  @ApiOperation(value = "", notes = """Fetch all projects.Support filtering by one or more attributes:
```
ajordens: Worth a newline between `projects.` and `Supports`

```
Fetch all projects.

Supports filtering by one or more attributes:
...
```

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.groovy
```diff
@@ -122,6 +127,51 @@ public class ProjectsController {
...
+    items.sort { Project a, Project b ->
+      return SearchUtils.score(b, attributes) - SearchUtils.score(a, attributes)
+    }
+
+  }
+
+  def fuzzyPredicate = { Project project, String key, String value  ->
+    boolean matches = SearchUtils.matchesIgnoreCase(project.properties, key, value)
+    if (!matches) {
+      if (project.config.pipelineConfigs.find { SearchUtils.matchesIgnoreCase(it.properties, key, value) } ) {
```
ajordens: I suspect that looking into `pipelineConfigs` is probably not providing any value

jeyrschabu:
made suggested changes. merging down

PR #164, opened 2016-11-30
Adds service account DAO health indicator

ttomsu:
@duftler @lwander @jtk54 PTAL


ttomsu:
Yes, until Greg (or I) get to https://github.com/spinnaker/spinnaker/issues/1018

PR #165, opened 2016-12-05
Adds more threads to S3 pool.

ttomsu:
@cfieber @duftler @ewiseblatt PTAL

@gardleopard FYI

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
@@ -134,27 +134,27 @@ public class S3Config {
 
   @Bean
   public ApplicationPermissionDAO applicationPermissionDAO(StorageService storageService) {
-    return new DefaultApplicationPermissionDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(1)), 45000);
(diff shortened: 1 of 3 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -134,17 +134,17 @@ public ApplicationDAO applicationDAO(StorageService storageService) {
 
   @Bean
   public ApplicationPermissionDAO applicationPermissionDAO(StorageService storageService) {
-    return new DefaultApplicationPermissionDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(1)), 45000);
+    return new DefaultApplicationPermissionDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(20)), 45000);
   }
```
ewiseblatt: If having a lot is likely, please add them to google-gcs config as well.
They dont need to match exactly but this PR looks like you advise a lot of threads here and below. Especially for these that might be in a more critical path than th others.

ttomsu: Made everything at least 20 threads.

PR #167, opened 2016-12-13
AuthorizationSupport conditional on FiatPermissionEvaluator

cfieber:
in the case where `fiat.autoConfigure = false` the `FiatPermissionEvaluator` bean is not present and this component causes startup errors


cfieber:
I don't actually see AuthorizationSupport used anywhere - is the better fix to just delete the class?

cfieber:
@ttomsu PTAL

cfieber:
this won't work, didn't see the usage in the annotations

PR #169, opened 2016-12-13
fix health indicators for ItemDAOs

cfieber:
increased health check timeout to 90s and fail if a poll interval larger than that is requested
instrumented StorageServiceSupport to add
`storageServiceSupport.cacheAge` gauge
`storageServiceSupport.cacheRefreshTime` timer


ajordens:
LGTM

PR #170, opened 2016-12-14
Refined metrics around cache maintainence.

ewiseblatt:
@cfieber 

I added more instrumentation around front50 cache management because it is a sensitivity point for performance and critical point for the accumulation of knowledge.

I am double counting refreshes only because it is easier that way. All refreshes use your original timer (I changed the implementation a little to be consistent with the others but am happy to change it back. Originally I made the change because I wasnt seeing the objectType breakout but that turned out to be a different issue. In simplifying that, the recorded time is going to be slightly less accurate because it includes updating the atomic reference. However given the latency of network calls and insignificant critical section on that lock, the difference is going to be negligable.

The second set of counts distinguishes scheduled refreshes vs spontaneous refreshes.
The double counting is because the refresh() API is public so it needs to be there anyway unless I start adding new methods. The public ones, if you cared for some reason, would be (total - scheduled - auto)

I also added counters on individual changes to see the rates of changes of resources themselves. I think this might be useful for a variety of reasons. I assume it is ok to do and will not cause contention interfering with the parallel stream processing.

Finally, I added some GCS metrics because our client library doesnt provide any.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -15,7 +15,9 @@
  */
 package com.netflix.spinnaker.front50.model;
 
+import com.netflix.spectator.api.Counter;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.Timer;
 import com.netflix.spinnaker.front50.exception.NotFoundException;
 import com.netflix.spinnaker.front50.support.ClosureHelper;
 import com.netflix.spinnaker.hystrix.SimpleHystrixCommand;
@@ -32,6 +34,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.function.ToDoubleFunction;
 import java.util.stream.Collectors;
 
 
@@ -45,6 +48,12 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
   private final Scheduler scheduler;
   private final int refreshIntervalMs;
   private final Registry registry;
+  private final Timer cacheRefreshTimer;      // All refreshes
+  private final Timer autoRefreshTimer;       // Only spontaneous refreshes in all()
+  private final Timer scheduledRefreshTimer;  // Only refreshes from scheduler
+  private final Counter addCounter;      // Newly discovered files during refresh
+  private final Counter removeCounter;   // Deletes discovered during refresh
+  private final Counter updateCounter;   // Updates discovered during refresh
 
   private final AtomicLong lastRefreshedTime = new AtomicLong();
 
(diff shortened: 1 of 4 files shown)


ewiseblatt:
@ttomsu

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -60,9 +69,32 @@ public StorageServiceSupport(ObjectType objectType,
     if (refreshIntervalMs >= HEALTH_MILLIS) {
       throw new IllegalArgumentException("Cache refresh time must be more frequent than cache health timeout");
     }
+    String typeName = objectType.name();
     this.registry = registry;
+
+    this.cacheRefreshTimer = registry.timer(
+      registry.createId("storageServiceSupport.cacheRefreshTime", "objectType", typeName));
```
ttomsu: should these metrics be prefixed with the component, i.e. "front50.storageServiceSupport.cacheRefreshTime" ? Is that a pattern followed by other components?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -52,17 +53,18 @@
...
-  private static int PIPELINE_STRATEGY_REFRESH_MS = (int) TimeUnit.MINUTES.toMillis(1);
-  private static int SERVICE_ACCOUNT_REFRESH_MS = (int) TimeUnit.MINUTES.toMillis(1);
+  private static int APPLICATION_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(15);
+  private static int APPLICATION_PERMISSIONS_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(45);
+  private static int PROJECT_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(30);
+  private static int NOTIFICATION_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(30);
+  private static int PIPELINE_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(10);
+  private static int PIPELINE_STRATEGY_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(20);
+  private static int SERVICE_ACCOUNT_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(30);
+  private static int SNAPSHOT_REFRESH_MS = (int) TimeUnit.SECONDS.toMillis(60);
```
ttomsu: If we're trying to stay in sync with S3, these values should probably be constants in a common module.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
ewiseblatt: The front50 is added externally. It isnt not part of the spectator metric.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -159,13 +199,16 @@ public void bulkImport(Collection<T> items) {
    * Update local cache with any recently modified items.
    */
   protected void refresh() {
-    long startTime = System.nanoTime();
-    allItemsCache.set(fetchAllItems(allItemsCache.get()));
-    long elapsed = System.nanoTime() - startTime;
+    long startTime = System.currentTimeMillis();
```
cfieber: switch these to back nanoTime for doing elapsed time (monotonic clock)

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -159,13 +199,16 @@ public void bulkImport(Collection<T> items) {
...
 
-    registry.timer("storageServiceSupport.cacheRefreshTime", "objectType", objectType.name()).record(elapsed, TimeUnit.NANOSECONDS);
+    cacheRefreshTimer.record(new Runnable() {
+      public void run() {
+          allItemsCache.set(fetchAllItems(allItemsCache.get()));
```
cfieber: any reason to create an inner runnable around this when we are already tracking the elapsed time (use .record(long instead of .record(Runnable) ?)

cfieber:
minor nits but LGTM

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
ewiseblatt: Originally I moved it because I was using datadog and didnt see the objectType metric anymore so thought this API might not have been working right, but then when it still didnt work, I realized it was because datadog converts mixed case labels to lower case labels so that's why it wasnt seeing objectType (because it converted it to objecttype). I then left it because it was consistent with everything else. I copied them in from a pending [working] PR I had before you submitted this (and had used a different label so didnt notice the mixed case thing).

I'll change all my uses into the pattern you had. I didnt realize this was even a usage pattern originally.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
ewiseblatt: Good catch. I didnt notice that I'd changed that.

ewiseblatt:
@cfieber 
PTAL, I changed my usage pattern from the runnable wrapper to tracking the time myself as you had done.

PR #175, opened 2016-12-20
Allow configurable S3 endpoint

viglesiasce:
This allows alternative S3 implementations such as Minio to be used with Front50.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
@@ -7,6 +7,7 @@ import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3Client;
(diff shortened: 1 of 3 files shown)


viglesiasce:
Thanks @cfieber!!

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -71,10 +72,15 @@ public AmazonS3 awsS3Client(AWSCredentialsProvider awsCredentialsProvider, S3Pro
     }
 
     AmazonS3Client client = new AmazonS3Client(awsCredentialsProvider, clientConfiguration);
-    Optional.ofNullable(s3Properties.getRegion())
-      .map(Regions::fromName)
-      .map(Region::getRegion)
-      .ifPresent(client::setRegion);
+    if (s3Properties.getEndpoint() != null){
```
ajordens: Minor code formatting ... (spacing between brackets)

```
if (s3Properties.getEndpoint != null) {
...
} else { 
...
}
```

If you could fix that, I'll get this merged in.

PR #178, opened 2016-12-22
Added object init to StorageService.

ewiseblatt:
@ttomsu FYI

@ajordens please review.
The main point is adding a hook to the StorageService interface to initialize a DAO so I can avoid logging errors that were causing concern for some customers while still keeping the errors should the files be deleted during execution.

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
@@ -433,7 +433,6 @@ public class GcsStorageService implements StorageService {
       try {
           timeExecute(patchTimer, obj_api.patch(bucketName, object.getName(), object));
       } catch (HttpResponseException e) {
-          log.error("writeLastModified failed to update {}\n{}", timestamp_path, e.toString());
           if (e.getStatusCode() == 404 || e.getStatusCode() == 400) {
               byte[] bytes = "{}".getBytes();
               ByteArrayContent content = new ByteArrayContent("application/json", bytes);
@@ -442,10 +441,14 @@ public class GcsStorageService implements StorageService {
                 log.info("Attempting to add {}", timestamp_path);
                 timeExecute(insertTimer, obj_api.insert(bucketName, object, content));
               } catch (IOException ioex) {
-                log.error("writeLastModified insert failed too: {}", ioex);
+                  log.error("writeLastModified failed to update {}\n{}",
+                            timestamp_path, e.toString());
+                  log.error("writeLastModified insert failed too: {}", ioex);
                 throw new IllegalStateException(e);
               }
           } else {
+              log.error("writeLastModified failed to update {}\n{}",
+                        timestamp_path, e.toString());
               throw new IllegalStateException(e);
           }
       } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -206,6 +206,14 @@ public T call() throws IOException {
   }
 
   /**
+   * Create a modification file if needed to avoid logging errors later.
+   */
+  public void initObjectType(ObjectType objectType) {
+      log.info("Initializing {}", objectType.group);
+      getLastModifiedHelper(objectType, false);
```
ajordens: Seems like an oddly named method to me ... with an unclear side-effect that a last modified file will be created.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -433,7 +441,10 @@ private void writeLastModified(String daoTypeName) {
       try {
           timeExecute(patchTimer, obj_api.patch(bucketName, object.getName(), object));
       } catch (HttpResponseException e) {
-          log.error("writeLastModified failed to update {}\n{}", timestamp_path, e.toString());
+          if (expected) {
```
ajordens: It would always seem like an error to me if this write failed ... not sure it's necessary to gate w/ `expected`?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -499,15 +510,24 @@ private void purgeOldVersions(String path) throws Exception {
...
+  private long getLastModifiedHelper(ObjectType objectType, boolean expected) {
       String path = daoRoot(objectType.group) + '/' + LAST_MODIFIED_FILENAME;
       try {
           return obj_api.get(bucketName, path).execute().getUpdated().getValue();
       } catch (HttpResponseException e) {
           if (e.getStatusCode() == 404) {
-              log.error("No timestamp file at {}. Creating a new one.", path);
-              writeLastModified(objectType.group);
+              if (expected) {
+                  log.error("No timestamp file at {}. Creating a new one.", path);
```
ajordens: Not sure I'd consider a missing last modified file to be an error ... but don't know the whole backstory that's motivating these changes.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -499,15 +510,24 @@ private void purgeOldVersions(String path) throws Exception {
...
+              } else {
+                  log.info("No timestamp file at {}. Creating a new one.", path);
+              }
+              writeLastModified(objectType.group, expected);
               return 0L;
```
ajordens: Not sure what the `0L` response is meant to indicate ... would seem like it should either be the last modified time if a file was written ... or an exception if the write failed?

ewiseblatt:
Redid this. Simplified it by just taking the message from error to info.

We have instrumentation around the cache sizes and I have refinements around adds and deletes so if someone were to delete the directory, those gauges and counters would pick up on it. The timestamp errors in the log file were another signal into that before those metrics were in place. It's still a little squirely to write some errors when the creates fail, but those errors are tolerated rather than propagated up since the timestamp is just an optimization.

ajordens:
LGTM

PR #179, opened 2017-01-05
Base front50 halconfig

lwander:
@duftler @jtk54 PTAL

### DIFF
--- a/halconfig/front50.yml
+++ b/halconfig/front50.yml
@@ -1,0 +1,14 @@
+server:
+  port: ${services.front50.port:8080}
+  address: ${services.front50.host:localhost}
+
+hystrix:
+  command:
+    default.execution.isolation.thread.timeoutInMilliseconds: 15000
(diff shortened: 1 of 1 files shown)


halconfig/front50.yml
```diff
@@ -0,0 +1,19 @@
...
+    default.execution.isolation.thread.timeoutInMilliseconds: 15000
+  threadpool:
+    DefaultNotificationDAO:
+      coreSize: 25
+      maxQueueSize: 100
+    DefaultPipelineDAO:
+      coreSize: 25
+      maxQueueSize: 100
+
+cassandra:
```
lwander: With our deprecation of cassandra is it OK to leave this out? @skim1420 @duftler

skim1420: Yes, we are moving to formally deprecate cassandra support.

jtk54: +1

PR #189, opened 2017-02-07
Removing unused loadObjectVersion storage API

robzienert:
No sense in having an API we don't use.

@spinnaker/reviewers @scotmoor  PTAL


lwander:
How do we generate pipeline diffs without using this endpoint?

robzienert:
A quick perusal - looks like front50 just uses `listObjectVersions`, then the diff is generated in deck.

ewiseblatt:
LGTM.

@lars none of the providers implemented it. The API itself for it wasnt well defined because it takes a VersionId parameter, which is storage service specific but thereI think there is no API to map versions into version id's to get a version id to pass in. So I think it was flawed in its present form anyway.

PR #191, opened 2017-02-12
(google) Add retries around GCS calls.

duftler:


duftler:
@lwander or @ewiseblatt please review.

duftler:
Depends on https://github.com/spinnaker/spinnaker-dependencies/pull/70.

duftler:
I focused on adding retries to the retrievals so a temporary failure (like a timeout or a network issue) doesn't cause an upstream operation to fail (e.g. a socket timeout while retrieving a child pipeline should not cause the parent pipeline to fail).

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -53,6 +54,18 @@
 @ConditionalOnExpression("${spinnaker.gcs.enabled:false}")
 @EnableConfigurationProperties(GcsProperties.class)
 public class GcsConfig {
+  @Value("${spinnaker.gcs.safeRetryMaxWaitIntervalMs:60000}")
+  Long maxWaitInterval;
+
+  @Value("${spinnaker.gcs.safeRetryRetryIntervalBaseSec:2}")
```
lwander: Why not put all of these under a `safeRetry` block? `safeRetryRetry...` reads kind of funny

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -257,25 +284,32 @@ public boolean supportsVersioning() {
...
-  public <T extends Timestamped> T
-  loadObject(ObjectType objectType, String objectKey)
-         throws NotFoundException {
+  public <T extends Timestamped> T loadObject(ObjectType objectType, String objectKey) throws NotFoundException {
     String path = keyToPath(objectKey, objectType.group);
     try {
-      StorageObject storageObject = timeExecute(loadTimer, obj_api.get(bucketName, path));
-      T item = deserialize(storageObject, (Class<T>) objectType.clazz, true);
-      item.setLastModified(storageObject.getUpdated().getValue());
+      StorageObject[] storageObjectHolder = new StorageObject[1];
```
lwander: What does this accomplish over writing `StorageObject storageObject = null`, and then assigning to that? Oh - does this have to be effectively final?

duftler: Yup.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
duftler: Good point. Done.

front50-gcs/front50-gcs.gradle
```diff
@@ -19,6 +19,7 @@ dependencies {
   compile 'io.reactivex:rxjava:1.0.10'
 
   spinnaker.group('google')
+  compile spinnaker.dependency('clouddriverGoogleCommon')
   compile spinnaker.dependency("korkSecurity")
 
   /* spinnaker.group('googleStorage') */
```
duftler: Turns out that dependency `googleStorage` is already in the `google` group, which is included just above. So I've dropped the line entirely.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
lwander: gr8

PR #197, opened 2017-02-14
try to reconcile dependencies

robfletcher:
front50-nflx is pulling in a newer version of AWS SDK transitively and breaking Front50 on startup. I'm trying to upgrade to the newer spinnaker-dependencies version that will bring in the same AWS version here but it also brings in a newer version of `clouddriver-google-common`.

Also a few things weren't using spinnaker-dependencies when it seems like they should.

### DIFF
--- a/front50-s3/front50-s3.gradle
+++ b/front50-s3/front50-s3.gradle
@@ -17,7 +17,7 @@
 dependencies {
   compile project(":front50-core")
 
-  compile 'com.netflix.spinnaker.clouddriver:clouddriver-aws:1.488.0'
+  compile spinnaker.dependency("clouddriverAws")
 //  compile 'com.aestasit.infrastructure.sshoogr:sshoogr:0.9.15'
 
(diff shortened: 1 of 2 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -592,15 +581,15 @@ public void doRetry(Closure operation,
...
-                           null,
                            Arrays.asList(500),
                            null,
                            maxWaitInterval,
                            retryIntervalBase,
                            jitterMultiplier,
-                           maxRetries);
+                           maxRetries,
+                           Collections.singletonMap("action", action),
+                           registry);
```
robfletcher: @duftler I think you made some changes here recently. Just want to make sure what I've done here is right.

robfletcher:
PTAL @spinnaker/google-reviewers

PR #199, opened 2017-02-16
Remove rootFolder from AZS config

EricJizbaMSFT:
For a few reasons:
1. Simplify the config
2. Match the folder structure for other providers, which are only 2 levels deep not 3 (in front50.yml, s3.bucket == azs.storageAccountName and s3.rootFolder == azs.storageContainerName)
3. Fix a bug where the 'last_modified' files were stored at the container level and everything else was stored at the (now removed) rootFolder level. Now everything is at the container level.

### DIFF
--- a/front50-azure/src/main/java/com/netflix/spinnaker/front50/config/AzureStorageConfig.java
+++ b/front50-azure/src/main/java/com/netflix/spinnaker/front50/config/AzureStorageConfig.java
@@ -64,10 +64,7 @@ public class AzureStorageConfig {
   @Bean
   public AzureStorageService azureStorageService(AzureStorageProperties azureStorageProperties) {
     // This is where we create the service
-    return new AzureStorageService(azureStorageProperties.getStorageConnectionString(),
-      azureStorageProperties.getStorageAccountName(),
-      azureStorageProperties.getStorageContainerName(),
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
@scotmoor @rguthriemsft @duftler @ewiseblatt PTAL

We realize this would be a breaking change for existing users of azure storage, but want to fix while it's new and before 1.0. When this and it's companion spinnaker PR https://github.com/spinnaker/spinnaker/pull/1432 are released, I will post in slack telling any possible users to follow the migration steps so that their data is not lost (download all metadata, update & restart front50, upload new metadata)

EricJizbaMSFT:
The naming between providers is confusing, so I'll give an example. Here is a path for an application object before this change:
spinnakervpqhwmi3ntc2e/spinnaker/front50/applications/sample/application-metadata.json

And here is the path after my change:
spinnakervpqhwmi3ntc2e/front50/applications/sample/application-metadata.json

I should've clarified about the levels. I was not counting "type/object container/object data" since that stuff is cloud agnostic (i.e. "applications/sample/application-metadata.json")

ewiseblatt:
Thanks. Makes total sense now.

ewiseblatt:
except I missed where the "sample" directory is when I was looking at your
PR implementation.

PR #207, opened 2017-03-01
feat(s3): Create specified bucket when it doesn't exist

lwander:
If you prefer `ensureBucketExists` does not get called during init that's OK. I filled out that method primarily for Halyard's consumption.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
@@ -16,16 +16,10 @@
 
 package com.netflix.spinnaker.front50.model;
 
+import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.model.AmazonS3Exception;
-import com.amazonaws.services.s3.model.GetObjectRequest;
-import com.amazonaws.services.s3.model.ListObjectsRequest;
-import com.amazonaws.services.s3.model.ListVersionsRequest;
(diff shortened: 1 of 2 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
```diff
@@ -50,22 +40,35 @@
...
     this.bucket = bucket;
     this.rootFolder = rootFolder;
     this.readOnlyMode = readOnlyMode;
+    this.region = region;
   }
 
   @Override
   public void ensureBucketExists() {
-    // do nothing
+    List<Bucket> buckets = amazonS3.listBuckets();
```
robzienert: [headBucket](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#headBucket-com.amazonaws.services.s3.model.HeadBucketRequest-) is probably a better fit here than iterating all existing buckets.

lwander: Oh neat, I was looking for something like that. Thanks

lwander: Updated

robzienert:
👍

tomaslin:
Please use naming conventions http://www.spinnaker.io/v1.0/docs/how-to-submit-a-patch

lwander:
Naming conventionalized

robzienert:
Wanna rebase & we'll merge this guy?

lwander:
We can get the same effect with `Update branch` + `Squash and merge`!

PR #209, opened 2017-03-01
Fixing a bug where search returned everything even on no match

jeyrschabu:
- adjusted search to use find and not collect on the test
 - also ensuring that passed queries are non-null


jeyrschabu:
@anotherchrisberry PTAL

PR #213, opened 2017-03-09
chore(fiat): Bumps Fiat API, adding EnableFiatAutoConfig annotation

ttomsu:
This is an example of what needs to be done in the rest of the non-Gate components that utilize the Fiat API module - basically a version bump + adding `@EnableFiatAutoConfig`.

@cfieber @jtk54 - PTAL.

Tests won't pass until spinnaker/fiat#154 is submitted and released.


ttomsu:
Tests back to green.

PR #221, opened 2017-04-21
chore(cassandra): Removing cassandra support

robzienert:
⚠️ DO NOT MERGE ⚠️ 

Removing Cassandra from the codebase. Merging should be held off until we've given sufficient communication to the community that this is coming down the pipes.

zqhzb6bo5fgoe

@spinnaker/reviewers PTAL


robzienert:
¯\_(ツ)_/¯ I feel like that should disable the block button. ಠ_ಠ

ewiseblatt:
There are changes to the spinnaker repository that should happen first as
well.
I can propose those in the next few days.
They should ultimately be merged before this because they are still going
to default front50 to cassandra.

robzienert:
Yeah, there's also additional cleanup on echo, scheduled-actions and... other internal services.

ewiseblatt:
What I'm suggesting is that we can take it out of the public configuration
yaml interface first so new users wont have access.
If we remove from spinnaker/config as a first step, which actually we can
do now, then existing configs could still have it, and the code will be
backing it, but new users wont see it as an option. I have a script
"change_cassandra" that we use in our install/config turn it off. That can
be used to turn it on. So if there is a platform that still needs it, they
can turn it on rather than having everyone else turn it off. It's built
into InstallSpinnaker, so if they are using that script then it's just
parameter.

robzienert:
That sounds good. How long do you propose keeping the ability to switch it
back on? Couple / few months?

Cassandra paths aren't getting updates (ex: pipeline templates won't get
support for Cass), so eventually it'll be far enough detached that it'll be
a liability for people using it. Hoping to avoid that scenario.

PR #222, opened 2017-04-25
feat(core): Adding pipeline template storage

robzienert:
The whole Redis `findById` iterating and filtering `all()` pattern feels wrong, and the `scope` interaction seems janky, but interested in initial feedback.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ObjectType.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ObjectType.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.front50.model;
 import com.netflix.spinnaker.front50.model.application.Application;
 import com.netflix.spinnaker.front50.model.notification.Notification;
 import com.netflix.spinnaker.front50.model.pipeline.Pipeline;
+import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplate;
 import com.netflix.spinnaker.front50.model.project.Project;
 import com.netflix.spinnaker.front50.model.serviceaccount.ServiceAccount;
 import com.netflix.spinnaker.front50.model.snapshot.Snapshot;
@@ -28,6 +29,7 @@ public enum ObjectType {
   PROJECT(Project.class, "projects", "project-metadata.json"),
   PIPELINE(Pipeline.class, "pipelines", "pipeline-metadata.json"),
   STRATEGY(Pipeline.class, "pipeline-strategies", "pipeline-strategy-metadata.json"),
+  PIPELINE_TEMPLATE(PipelineTemplate.class, "pipeline-templates", "pipeline-template-metadata.json"),
   NOTIFICATION(Notification.class, "notifications", "notification-metadata.json"),
   SERVICE_ACCOUNT(ServiceAccount.class, "serviceAccounts", "serviceAccount-metadata.json"),
   APPLICATION(Application.class, "applications", "application-metadata.json"),
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.front50.model.pipeline;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.front50.model.Timestamped;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTemplate extends HashMap<String, Object> implements Timestamped {
+
+  @JsonIgnore
+  @SuppressWarnings("unchecked")
+  public List<String> getScopes() {
+    Map<String, Object> metadata = (Map<String, Object>) super.get("metadata");
+    if (metadata == null || metadata.isEmpty()) {
+      return Collections.emptyList();
+    }
+    return (List<String>) metadata.get("scopes");
+  }
+
+  @Override
+  public String getId() {
+    return (String) super.get("id");
+  }
+
+  @Override
+  public Long getLastModified() {
+    String updateTs = (String) super.get("updateTs");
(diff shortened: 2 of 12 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTemplate extends HashMap<String, Object> implements Timestamped {
+
+  @JsonIgnore
+  @SuppressWarnings("unchecked")
+  public List<String> getScope() {
+    Map<String, Object> metadata = (Map<String, Object>) super.get("metadata");
+    if (metadata.isEmpty()) {
```
ajordens: empty or null maybe?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTemplate extends HashMap<String, Object> implements Timestamped {
+
+  @JsonIgnore
+  @SuppressWarnings("unchecked")
+  public List<String> getScope() {
```
ajordens: `scope` vs `scopes`? 

I suspect there's some motivation to the naming ... I just momentarily got caught up `scope` implying singular and it returning a list.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public void setLastModifiedBy(String lastModifiedBy) {
+    super.put("lastModifiedBy", lastModifiedBy);
+  }
+
+  public boolean containsAnyScope(List<String> scope) {
+    for (String s : scope) {
+      if (getScope().contains(s)) {
```
ajordens: Any worry about case sensitivity wrt to scopes?

front50-gcs/src/main/java/com/netflix/spinnaker/front50/config/GcsConfig.java
```diff
@@ -166,6 +169,11 @@ public PipelineDAO pipelineDAO(GcsStorageService service, Registry registry) {
   }
 
   @Bean
+  public PipelineTemplateDAO pipelineTemplateDAO(StorageService storageService, Registry registry) {
+    return new DefaultPipelineTemplateDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(25)), PIPELINE_TEMPLATE_REFRESH_MS, registry);
```
ajordens: Perhaps too many threads (depends on how many # of pipeline templates and concurrent calls to getByScope) but I've got a PR in flight that lets this be a little more configurable.

front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
```diff
@@ -138,6 +140,11 @@ public PipelineDAO pipelineDAO(StorageService storageService, Registry registry)
   }
 
   @Bean
+  public PipelineTemplateDAO pipelineTemplateDAO(StorageService storageService, Registry registry) {
+    return new DefaultPipelineTemplateDAO(storageService, Schedulers.from(Executors.newFixedThreadPool(25)), 10000, registry);
```
ajordens: Technically you've got a different refresh interval between S3 and GCS ... my PR will clean this up as well but I'd suggest dropping it down to 60s.

Any call to `all()` will do an on-demand refresh of any deltas anyways so it's more about having a reasonably fresh cache in the event of a hystrix fallback.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
```diff
@@ -0,0 +1,113 @@
...
+      pipelineTemplateDAO.findById(id);
+    } catch (NotFoundException e) {
+      return;
+    }
+    throw new DuplicatePipelineTemplateIdException("A pipeline template with the id " + id + " already exists");
+  }
+
+  @ExceptionHandler(InvalidPipelineTemplateRequestException.class)
+  @ResponseStatus(HttpStatus.BAD_REQUEST)
+  Map handleInvalidPipelineTemplateRequestException(InvalidPipelineTemplateRequestException e) {
```
ajordens: Nothing to do with this PR but I reckon we can do a much better job of exception handling that doesn't require per-controller exception classes ...

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.front50.model.Timestamped;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTemplate extends HashMap<String, Object> implements Timestamped {
```
ajordens: Doesn't have to `extends HashMap<>` if it makes more sense to do a concrete object w/ @JsonAnyGetter/JsonAnySetter for the dynamic pipeline bits.

Would let you concretely model the `metadata` but meh?

ajordens:
I don't see anything out of the ordinary, lgtm (few comments above but nothing blocking I don't think).

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
robzienert: I debated on this a little bit, then decided to just throw something out there. I suspect that most people will scope templates to a single item (global or a specific app), which is why I chose the singular. Definitely happy to make it plural, since it is a list.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
robzienert: Agree, I'm still getting my bearings on how we do error handling in general in the services. I don't want to run off doing something different until I've got a good handle on how to do it better.

ajordens: I think it's completely reasonable to do it this way in front50 for now ... and then one of us can do a collective refactor across the board when we've got a more reasonable pattenr.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ObjectType.java
```diff
@@ -28,6 +29,7 @@
   PROJECT(Project.class, "projects", "project-metadata.json"),
   PIPELINE(Pipeline.class, "pipelines", "pipeline-metadata.json"),
   STRATEGY(Pipeline.class, "pipeline-strategies", "pipeline-strategy-metadata.json"),
+  PIPELINE_TEMPLATE(PipelineTemplate.class, "pipelineTemplates", "pipeline-template-metadata.json"),
```
ajordens: maybe `pipeline-templates` vs `pipelineTemplates` to be consistent with the naming convention of other types.

image

PR #230, opened 2017-05-11
feat(oraclebmcs): Add Oracle BMCS Storage Service provider

simonlord:
Note that we manually download our SDK and add to classpath in our gradle build script. We did this in clouddriver too. This will be fixed once
our SDK is available on Maven central - soon(ish).

Also note that due to the Front50 classpath containing Jersey 1 (brought in by kork IIRC?) we've had to not use our SDK's Jersey 2 based storage
client and instead create a minimal Jersey 1 implementation. Hopefully this can be improved upon in the future either by bumping kork to use
Jersey 2?

### DIFF
--- a/front50-oracle-bmcs/src/main/java/com/netflix/spinnaker/front50/config/OracleBMCSProperties.java
+++ b/front50-oracle-bmcs/src/main/java/com/netflix/spinnaker/front50/config/OracleBMCSProperties.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2017 Oracle America, Inc.
+ *
+ * The contents of this file are subject to the Apache License Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.front50.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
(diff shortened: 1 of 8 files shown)


front50-oracle-bmcs/src/main/java/com/netflix/spinnaker/front50/config/OracleBMCSProperties.java
```diff
@@ -0,0 +1,89 @@
...
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.front50.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("spinnaker.oraclebmcs")
+public class OracleBMCSProperties {
+
+  private String bucketName = "_spinnaker_front50_data";
```
lwander: I know for S3 & GCS this has to be globally unique, is this the case here? If so, this default might confuse people

simonlord: I'm not sure i fully understand. However, the Oracle BMCS Object Storage Service is a regional service, not global and the buckets names need only be unique within the tenancy/account. Does that help?

lwander: Ah that makes sense, so it shouldn't cause any trouble, sounds good.

lwander:
Build & clean looks good

PR #231, opened 2017-05-18
feat(fiat): Updates to new read-only permission model

ttomsu:
@jtk54 @cfieber PTAL


jtk54:
After unit tests pass.

ttomsu:
Psht - unit tests? Who needs 'em...

PR #233, opened 2017-05-22
fix(cassandra): Disable Pipeline Template storage capability

robzienert:
Cassandra is deprecated. Will not receive pipeline template support.

@spinnaker/reviewers PTAL

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.front50.controllers;
 
+import com.netflix.spinnaker.front50.exception.BadRequestException;
 import com.netflix.spinnaker.front50.exception.NotFoundException;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplate;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplateDAO;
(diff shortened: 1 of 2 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
```diff
@@ -37,29 +37,42 @@
...
-  @Autowired
-  PipelineTemplateDAO pipelineTemplateDAO;
+  @Autowired(required = false)
+  PipelineTemplateDAO pipelineTemplateDAO = null;
 
   // TODO rz - Add fiat authz
 
   @RequestMapping(value = "", method = RequestMethod.GET)
   List<PipelineTemplate> list(@RequestParam(required = false, value = "scopes", defaultValue = "global") List<String> scopes) {
+    if (pipelineTemplateDAO == null) {
```
ajordens: Doesn't look like we have an `@ExceptionHandler` for it but we could add that and kill two birds with one stone.

(otherwise we don't get the exception message in the response)

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
```diff
@@ -37,29 +37,42 @@
...
+      throw PipelineTemplateFeatureUnsupportedException.build();
+    }
     return (List<PipelineTemplate>) pipelineTemplateDAO.getPipelineTemplatesByScope(scopes);
```
ajordens: could cut down the duplicated code w/ ...

```
PipelineTemplateDAO getPipelineTemplateDAO() {
   if (pipelineTemplateDAO == null) {
      throw exception;
   }

   return pipelineTemplateDAO;
}
```

and then just use `getPipelineTemplateDAO()` in each method.

ajordens: realize I didn't do this in the `EntityTagsController` 👎

PR #235, opened 2017-05-26
fix(entity_tags): Fix item cache refresh

ewiseblatt:
EntityTags are not refreshing in the cache. With this change they are.

I recall the heuristic for updating streams needing more than 1 thread.
I'm pretty sure it was starved of threads having only 1.

@ajordens


ajordens:
Ahh, good catch. I suspect it might be the two `subscribeOn` in `fetchAllItems()`.

Do you mind changing

```
@Canonical
  static class PerObjectType {
    int threadPool
    long refreshMs
  }
```

to

```
static class PerObjectType {
    int threadPool
    long refreshMs

    PerObjectType(int threadPool, long refreshMs) {
      setThreadPool(threadPool)
      setRefreshMs(refreshMs)
    }

    void setThreadPool(int threadPool) {
      if (threadPool <= 1) {
        throw new IllegalArgumentException("threadPool must be >= 1")
      }
      this.threadPool = threadPool
    }

    void setRefreshMs(long refreshMs) {
      this.refreshMs = refreshMs
    }
  }
```

That'll raise an exception if someone ever attempts to configure an incorrectly sized pool.

FWIW this didn't come up for us as the entity tags pool size is overridden in our `front50.yml`:

```
storageService:
  entityTags:
    threadPool: 50
    refreshMs: 120000
```

ajordens:
Otherwise LGTM

PR #238, opened 2017-05-30
fix(test): Remove account front redis provider to get tests to pass

ttomsu:


ttomsu:
Root cause is `@IgnoreIf({ RedisTestHelper.redisUnavailable() })` at the top of the file. It passes on Travis-CI because of this, but fails on developer machines that actually have `redis://localhost:6379` available. I don't think @anotherchrisberry had Redis available when he cut https://github.com/spinnaker/front50/pull/220

PR #239, opened 2017-05-30
fix(build): Set sudo to false for metal-Travis build

ttomsu:
Experimental PR.


ttomsu:
Looks like you need sudo to install java.

PR #244, opened 2017-07-10
fix(permissions): Merge permissions for single application config call

ttomsu:
@veeresh1982 please take for a spin

If successful, this will need to be cherry picked into the 1.1 branch


veeresh1982:
LGTM.
With this change, I was able to add permission and see it get displayed in UI. I did my fiat testing steps which I had done earlier for simple k8s and all results looks expected.

PR #248, opened 2017-07-14
feat(pipeline_template): delete template

danielpeach:
@robzienert I was going to put the check for dependent pipeline configs in Orca, but realized that wouldn't work for anyone using Fiat without read access to all applications.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
@@ -15,12 +15,17 @@
  */
 package com.netflix.spinnaker.front50.controllers;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import com.netflix.spinnaker.front50.exception.BadRequestException;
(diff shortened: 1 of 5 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
```diff
@@ -69,6 +80,41 @@ PipelineTemplate update(@PathVariable String id, @RequestBody PipelineTemplate p
...
   }
 
+  @RequestMapping(value = "{id}", method = RequestMethod.DELETE)
+  void delete(@PathVariable String id) {
+    checkForDependentConfigs(id);
+    getPipelineTemplateDAO().delete(id);
+  }
+
+  @VisibleForTesting
+  void checkForDependentConfigs(String templateId) {
```
robzienert: You'd also want to check for dependent templates, since templates can inherit from another.

robzienert: https://github.com/spinnaker/dcd-spec/blame/master/PIPELINE_TEMPLATES.md#L105

danielpeach: That is a good point.

PR #250, opened 2017-07-17
fix(pipeline_templates): don't NPE if template has no scopes

danielpeach:
@robzienert 

Someone hit this, and I think it's reasonable for a template to have no scopes.


robzienert:
👍

PR #260, opened 2017-08-09
fix(serviceAccounts): Delete service account front Fiat when deleted …

ttomsu:
…from Front50

cc @msloes


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.groovy
```diff
@@ -62,6 +62,11 @@ public class ServiceAccountsController {
   void deleteServiceAccount(@PathVariable String serviceAccountId) {
     def acct = serviceAccountDAO.findById(serviceAccountId)
     serviceAccountDAO.delete(serviceAccountId)
+    try {
+      fiatService.logoutUser(serviceAccountId)
+    } catch (RetrofitError re) {
+      log.warn("Could not delete service account user $serviceAccountId", re)
```
jtk54: What are the ramifications if we can't log the service account out?

jtk54: Is a warning sufficient for being unable to log the service account user out?

robzienert:
@ttomsu ping

PR #262, opened 2017-08-17
chore(orca): migrate stage timeouts and failure policies

robfletcher:
See https://github.com/spinnaker/orca/pull/1554

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/StageTimeoutMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/StageTimeoutMigration.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/StageTimeoutMigration.java
```diff
@@ -0,0 +1,69 @@
...
+    return getStages(pipeline)
+      .stream()
+      .anyMatch(stage -> stage.containsKey("stageTimeoutMs"));
+  }
+
+  private void migrate(ItemDAO<Pipeline> dao, Pipeline pipeline, String type) {
+    getStages(pipeline)
+      .stream()
+      .filter(stage -> stage.containsKey("stageTimeoutMs"))
+      .forEach(stage -> stage.put("timeoutMs", stage.get("stageTimeoutMs")));
```
robzienert: No interest in removing the `stageTimeoutMs` value?

robzienert: Ah, nvm... can't do that and have the migration backwards compat.

PR #266, opened 2017-08-28
feat(pipeline_template): Get dependent pipeline configs for template id

robzienert:
Foundation work for being able to perform plans on all dependent pipelines during a template update operation. Updates the dependent template work to be recursive as well.

@emjburns

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
@@ -21,11 +21,11 @@ import com.netflix.spinnaker.front50.exception.BadRequestException;
 import com.netflix.spinnaker.front50.exception.NotFoundException;
 import com.netflix.spinnaker.front50.exceptions.DuplicateEntityException;
 import com.netflix.spinnaker.front50.exceptions.InvalidRequestException;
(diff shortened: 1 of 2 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineTemplateController.java
```diff
@@ -105,29 +119,57 @@ void checkForDependentConfigs(String templateId) {
...
         + " depend on this template: " + String.join(", ", dependentConfigIds));
     }
   }
 
   @VisibleForTesting
-  void checkForDependentTemplates(String templateId) {
+  List<String> getDependentTemplates(String templateId) {
     List<String> dependentTemplateIds = new ArrayList<>();
-
     getPipelineTemplateDAO().all()
```
danielpeach: This might get real slow given the recursive call...

robzienert: That was one of my concerns as well. I think I'll make it optional via query param

danielpeach: I think you should be able to loop just twice over the set of templates. The first time you would build a map `templateId -> sourceId`. The second time, for each template, you would walk down the map (`A -> B -> C -> ...`) for each template until you reach a root template.

PR #269, opened 2017-08-30
fix(web): change refresh to default behavior

emjburns:
Reverting to default refresh behavior (true). 

@ajordens @ewiseblatt PTAL


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/DefaultPipelineDAO.java
```diff
@@ -39,7 +39,7 @@ public DefaultPipelineDAO(StorageService service,
 
   @Override
   public String getPipelineId(String application, String pipelineName) {
-    Pipeline matched = getPipelinesByApplication(application, false)
+    Pipeline matched = getPipelinesByApplication(application, true)
```
ewiseblatt: Why do you need to pass this at all since you have the default arg? I presume here you want the "default" more than the specific behavior?

emjburns: True that isn't necessary.

PR #270, opened 2017-08-30
feat(health): support differing intervals for item dao health

cfieber:
existing implementations remain on a 30 second interval with the exception of S3StorageService subclasses which will check on a 2 second interval.


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.groovy
```diff
@@ -1,43 +0,0 @@
-/*
```
cfieber: converted to java to add `default` method

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -172,6 +172,10 @@ public boolean isHealthy() {
     return (System.currentTimeMillis() - lastRefreshedTime.get()) < getHealthMillis() && allItemsCache.get() != null;
   }
 
+  public long getHealthIntervalMillis() {
```
cfieber: this feels a little dirty because it ends up overriding the `ItemDao.getHealthIntervalMillis` even though it doesn't directly implement ItemDAO (its subclasses do)

ajordens:
seems reasonable ... I can live with the `getHealthIntervalMillis()` override shenanigans.

PR #272, opened 2017-09-04
feat(s3): Allow to disable versioning via config

bacher09:
[Minio s3](https://github.com/minio/minio) compatible storage [isn't supporting versioning](https://github.com/minio/minio/issues/2111) and this causes some issues when using it as storage backend for front50. So, I think it's nice idea to make versioning configurable in configuration file, so if someone is using s3 compatible storage that isn't supporting this, it would be easy to disable it.

As an alternative option, will be autodetection of versioning support on storage, but this might not work well on many different s3 compatible storages.


bacher09:
@spinnaker/reviewers please take a look

ttomsu:
cc @gardleopard

PR #276, opened 2017-09-06
feat(aws): Support for incremental cache updates based on s3 events

ajordens:
Adds a new `ObjectKeyLoader` implementation that subscribes to an sqs
queue and receives events generated in response to s3 modifications.

These events should be received no more than 1s after the modification.

This PR also removes the unused `loadObjectsWithPrefix` StorageService
method.

### DIFF
--- a/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
+++ b/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
@@ -104,31 +104,6 @@ public class AzureStorageService implements StorageService {
   }
 
   @Override
-  public <T extends Timestamped> Collection<T> loadObjectsWithPrefix(ObjectType objectType, String prefix, int maxResults) {
-    Set<T> blobs = new HashSet<>();
-    String key = buildKeyPath(objectType.group, prefix, "");
-    try {
-      ResultContinuation token = null;
-      EnumSet<BlobListingDetails> listDetails = EnumSet.of(BlobListingDetails.METADATA);
-      do {
-        ResultSegment<ListBlobItem> result = blobContainer.listBlobsSegmented(key, true, listDetails, maxResults, token, null, null);
-        token = result.getContinuationToken();
-
-        for (ListBlobItem item : result.getResults()) {
-          blobs.add(deserialize((CloudBlockBlob)item, (Class<T>) objectType.clazz));
-        }
-      } while (token != null);
-    } catch (IOException e) {
-      throw new IllegalStateException("Unable to deserialize object(s) (key: " + key + ")", e);
-    } catch (StorageException se) {
-      logStorageException(se, key);
-    } catch (Exception e) {
-      log.error("Failed to retrieve objects with prefix {}: {}", key, e.getMessage());
-    }
-    return blobs;
-  }
-
-  @Override
   public void deleteObject(ObjectType objectType, String objectKey) {
     String key = buildKeyPath(objectType.group, objectKey, objectType.defaultMetadataFilename);
     try {
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/EventingS3ObjectKeyLoader.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/EventingS3ObjectKeyLoader.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
```diff
@@ -104,31 +104,6 @@ public boolean supportsVersioning() {
   }
 
   @Override
-  public <T extends Timestamped> Collection<T> loadObjectsWithPrefix(ObjectType objectType, String prefix, int maxResults) {
```
ajordens: This was never called by a controller method and has been removed (similar changes to other `StorageService` implementations).

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/EventingS3ObjectKeyLoader.java
```diff
@@ -0,0 +1,273 @@
...
+
+      log.debug(
+        "Received Event (objectType: {}, type: {}, key: {}, delta: {})",
+        keyWithObjectType.objectType,
+        eventType,
+        keyWithObjectType.key,
+        System.currentTimeMillis() - eventTime.getMillis()
+      );
+
+      objectKeysByLastModifiedCache.put(keyWithObjectType, eventTime.getMillis());
```
ajordens: Seemed easier to avoid concurrency issues if we tracked detected modifications and applied when `listObjectKeys()` was called ... vs trying to apply them directly to the cache as changes came in.

From testing I do not believe that the rate of change is going to be overwhelming.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java
```diff
@@ -94,29 +94,6 @@ public boolean supportsVersioning() {
   }
 
   @Override
-  public <T extends Timestamped> Collection<T> loadObjectsWithPrefix(ObjectType objectType, String prefix, int maxResults) {
-    ObjectListing bucketListing = amazonS3.listObjects(
-      new ListObjectsRequest(bucket, (buildTypedFolder(rootFolder, objectType.group) + "/" + prefix).toLowerCase(), null, null, maxResults)
-    );
-    List<S3ObjectSummary> summaries = bucketListing.getObjectSummaries();
-
-    // TODO-AJ this is naive and inefficient
```
ajordens: Good thing we were never actually calling it!

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/TemporarySQSQueue.java
```diff
@@ -0,0 +1,165 @@
...
+  void markMessageAsHandled(String receiptHandle) {
+    try {
+      amazonSQS.deleteMessage(temporaryQueue.sqsQueueUrl, receiptHandle);
+    } catch (ReceiptHandleIsInvalidException e) {
+      log.warn("Error deleting message, reason: {} (receiptHandle: {})", e.getMessage(), receiptHandle);
+    }
+  }
+
+  @PreDestroy
+  void shutdown() {
```
ajordens: Makes a best attempt to cleanup the queue and subscription.

We'll probably need to do something to detect orphaned queues and remove them periodically.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/TemporarySQSQueue.java
```diff
@@ -0,0 +1,165 @@
...
+      .map(Topic::getTopicArn)
+      .findFirst()
+      .orElseThrow(() -> new IllegalArgumentException("No SNS topic found (topicName: " + topicName + ")"));
+  }
+
+  private TemporaryQueue createQueue(String snsTopicArn, String sqsQueueArn, String sqsQueueName) {
+    String sqsQueueUrl = amazonSQS.createQueue(
+      new CreateQueueRequest()
+        .withQueueName(sqsQueueName)
+        .withAttributes(Collections.singletonMap("MessageRetentionPeriod", "60")) // 60s message retention
```
ajordens: purposely short retention period

ajordens:
Thus far I've been making changes directly to applications/pipelines and ensuring the cache reconciles properly within the expected amount of time. 

Need to do another round of live end-to-end testing of this to ensure nothing appears out of the ordinary in the UI.

front50-s3/src/main/java/com/netflix/spinnaker/front50/model/TemporarySQSQueue.java
cfieber: should be fairly easy to do with the instanceid in the queue name

PR #284, opened 2017-10-25
fix(web): create application null name error with Azure Storage

yungez:
To fix [create application failure with Azure Storage](https://github.com/spinnaker/spinnaker/issues/1869). @rguthriemsft @scotmoor @spinnaker/netflix-reviewers @ZhijunZhao @yaweiw

### DIFF
--- a/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
+++ b/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
@@ -95,9 +95,15 @@ public class AzureStorageService implements StorageService {
       CloudBlockBlob blob = getBlobContainer().getBlockBlobReference(key);
       if (blob.exists()) {
         return deserialize(blob, (Class<T>) objectType.clazz);
+      } else {
+        // if blob not exists, throw NotFoundException, to align with other cloud provider
+        throw new NotFoundException("Azure blob not exists: " + key);
       }
     } catch (StorageException se) {
       logStorageException(se, key);
+      if (se.getHttpStatusCode() == org.apache.http.HttpStatus.SC_NOT_FOUND) {
+        throw new NotFoundException("Azure blob not found: " + key);
+      }
     }
     catch (Exception e) {
       log.error("Failed to retrieve {} object: {}: {}",
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
@@ -117,13 +117,15 @@ public class ApplicationsController {
   @RequestMapping(method = RequestMethod.GET, value = "/{applicationName:.+}")
   Application get(@PathVariable final String applicationName) {
     def app = applicationDAO.findByName(applicationName.toUpperCase())
-    try {
-      def perm = applicationPermissionDAO?.findById(app.name)
-      if (perm?.permissions?.isRestricted()) {
-        app.details().put("permissions", perm.permissions)
+    if (app != null) {
+      try {
+        def perm = applicationPermissionDAO?.findById(app.name)
+        if (perm?.permissions?.isRestricted()) {
+          app.details().put("permissions", perm.permissions)
+        }
+      } catch (NotFoundException nfe) {
+        // ignored.
       }
-    } catch (NotFoundException nfe) {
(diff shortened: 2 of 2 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
```diff
@@ -117,13 +117,15 @@ public class ApplicationsController {
   @RequestMapping(method = RequestMethod.GET, value = "/{applicationName:.+}")
   Application get(@PathVariable final String applicationName) {
     def app = applicationDAO.findByName(applicationName.toUpperCase())
-    try {
-      def perm = applicationPermissionDAO?.findById(app.name)
-      if (perm?.permissions?.isRestricted()) {
-        app.details().put("permissions", perm.permissions)
+    if (app != null) {
```
milismsft: do you know what causes the app not to be found in applicationDAO?

yungez: not sure yet. while after the fixing, create application succeeded. Isn't the get before creating, so not found?

milismsft: You need to dig a bit deeper and understand how applicationDAO is built and if the answer it might be that is not fully supported for Azure. It looks like it might have to do with some permission setting so may be we are missing something during the Front50 configuration? 
And since this change is in common code you will have to make a compelling case when pushing the PR into main branch why adding this check is needed.

yungez: that make sense. Actually I'm looking at permission check support for azure part. For the null check itself, no matter which provider it is, if app is null, later call on app.name will fail with null object error.

yungez:
@ttomsu  Hello Travis, we saw there're permission support added in front50. Could you pls share expected behavior for cloud provider to go through creating new application flow after permission change? So that we could support Azure Storage. To share more context of this PR and this ask: we found that create new application flow with Azure Storage failed since this July, after investigation, it started failing since this commit, https://github.com/spinnaker/front50/commit/1b27f6d4ee7c90c6bea40be03f889c08f413c559#diff-89c9cd9a5a04b047cbbfd9d54f66018d. We're trying to fix the issue now, besides the simple null check, we'd like to understand what's expected behavior for cloud provider regarding permission, when creating new application. Thanks for your help. any other information need feel free to ping me.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
ttomsu: A shorter way to express this is in groovy: 
```
if (app && app.name)
```

yungez: thanks you @ttomsu for the comments. while besides directly fix this null error, we'd like to learn more about permission feature in spinnaker. what behavior should be added in cloud provider to support permission when creating application?

front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
```diff
@@ -95,9 +95,15 @@ public boolean supportsVersioning() {
       CloudBlockBlob blob = getBlobContainer().getBlockBlobReference(key);
       if (blob.exists()) {
         return deserialize(blob, (Class<T>) objectType.clazz);
+      } else {
```
ttomsu: No need for an `else` block here:
```
if (blob.exists()) {
  return ...
}
throw new NotFoundException...
```

ajordens:
Can you give this a try after spinnaker/front50#291 has merged.

That PR throws a couple more exceptions from `AzureStorageService.loadObject()` to bring it in line with the other providers. I do not believe the changes to `ApplicationsController` should be necessary.

yungez:
@ajordens, thanks for the info . actually this PR has same fixing on azure provider loadobject() not found behavior. pls see https://github.com/spinnaker/front50/pull/284/files change on AzureStorageService.java. it's ok either one is merged. the change on ApplicationController is just defensive code. I can close this PR then.

yungez:
close the PR since same fixing is merged https://github.com/spinnaker/front50/pull/291

PR #285, opened 2017-10-26
feat(keel): adding basic storage of intents

emjburns:
Basics of intent storage for keel. Object looks like https://github.com/spinnaker/keel/blob/master/README.md.

@ajordens @robzienert PTAL

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/IntentController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/IntentController.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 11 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/IntentController.java
```diff
@@ -0,0 +1,86 @@
...
+  private void checkForDuplicateIntents(String id) {
+    try {
+      getIntentDAO().findById(id);
+    } catch (NotFoundException e) {
+      return;
+    }
+    throw new DuplicateEntityException("An intent with the id " + id + " already exists");
+  }
+
+  private IntentDAO getIntentDAO() {
```
robzienert: I'd get rid of this accessor method, it's fine to reference the property directly.

emjburns: Done.

PR #287, opened 2017-10-28
fix(s3): don't create sqs and sns topics when eventing is disabled

tomaslin:
This leads to an exception when using s3 compatible systems like minio. Error creating bean with name 'awsSQSClient' defined in class path resource [com/netflix/spinnaker/front50/config/S3Config.class]


ajordens:
LGTM

PR #290, opened 2017-11-16
fix(intent, pipeline_template): controllers conditional on storage ba…

emjburns:
…ckend

@lwander @tomaslin this will fix the local front50 startup problem by disabling pipeline templates and intents in front50 if it's running locally. Running locally with no storage is not really supported (defaults to Cassandra in memory..)

@robzienert I've removed your error message to make the behavior consistent.


tomaslin:
confirming this fixes that evaluates to false issue 
LGTM

PR #293, opened 2017-11-29
fix(pipeline_templates): Fix template controller with GCS.

jtk54:
The `ConditionalOnBean` was causing the `PipelineTemplateController` not to be wired properly and the endpoints 404 from Orca/roer when configured with GCS as the backing store.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/config/CommonStorageServiceDAOConfig.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/config/CommonStorageServiceDAOConfig.java
@@ -44,10 +44,12 @@ import com.netflix.spinnaker.front50.model.tag.DefaultEntityTagsDAO;
 import com.netflix.spinnaker.front50.model.tag.EntityTagsDAO;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import rx.schedulers.Schedulers;
 
 import java.util.concurrent.Executors;
 
+@Configuration
 public class CommonStorageServiceDAOConfig {
   @Bean
   @ConditionalOnMissingBean(ObjectKeyLoader.class)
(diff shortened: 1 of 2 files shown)


emjburns:
Can you tell me more about the error you're seeing in the wiring? If the `PipelineTemplateController` isn't wired properly with `ConditionalOnBean`, was it working before? 

To revert back to the previous version (throwing an error if your storage backend isn't supported for pipeline templates) you'll need to revert all changes to `PipelineTemplateController` from this PR: https://github.com/spinnaker/front50/commit/1032fd22f639ac6a629a24bad727a76c94ce3eec

jtk54:
The controller behaved as expected with `(required = false)` on `PipelineTemplateDAO`. With the `ConditionalOnBean(PiplineTemplateDAO)` and no `(required = false)`, I see no exceptions on front50 startup but the endpoints 404 in Orca and roer when I try to save/update pipeline templates to GCS.

In this PR, if the storage backend doesn't support pipeline templates, the `PiplineTemplateDAO` autowiring fails and front50 does not start.

What is the intended behavior if the storage backend doesn't support pipeline templates?

emjburns:
The endpoints will 404 because `ConditionalOnBean` means that if the bean doesn't exist, the controller won't be there. I think the beans are getting wired in the wrong order, because pipeline templates are supported by GCS. Let me look into this some more. 

The behavior we want is for front50 to NOT need to support pipeline templates (or newer features that we don't add backwards compatibility for storage providers that are not supported - like redis and cassandra). So, if the storage backend doesn't support pipeline templates front50 should still start, but you won't be able to use the pipeline template endpoints (either because 404s are returned or a message saying 'not supported with your storage backend' is thrown: https://github.com/spinnaker/front50/commit/1032fd22f639ac6a629a24bad727a76c94ce3eec#diff-57a0647ae75cadbbcce893684e9869a2L196)

jtk54:
Talked offline - closing this PR in favor of a different refactor.

emjburns:
Fi in https://github.com/spinnaker/front50/pull/294

PR #302, opened 2018-03-02
fix(pipelineTemplates): Validate Pipelines with Templates

beckje01:
Adds validation for any pipeline with spinnaker managed pipeline templates. This will ensure the template is available on create or update.

https://github.com/spinnaker/spinnaker/issues/2492

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
@@ -16,12 +16,16 @@
 
 package com.netflix.spinnaker.front50.controllers
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.front50.exception.BadRequestException
+import com.netflix.spinnaker.front50.exception.NotFoundException
 import com.netflix.spinnaker.front50.exceptions.DuplicateEntityException
 import com.netflix.spinnaker.front50.exceptions.InvalidEntityException
 import com.netflix.spinnaker.front50.exceptions.InvalidRequestException
 import com.netflix.spinnaker.front50.model.pipeline.Pipeline
 import com.netflix.spinnaker.front50.model.pipeline.PipelineDAO
-
+import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplateDAO
+import com.netflix.spinnaker.front50.model.pipeline.TemplateConfiguration
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.security.access.prepost.PostFilter
 import org.springframework.security.access.prepost.PreAuthorize
@@ -32,6 +36,9 @@ import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
+import static com.netflix.spinnaker.front50.model.pipeline.Pipeline.TYPE_TEMPLATED
+import static com.netflix.spinnaker.front50.model.pipeline.TemplateConfiguration.TemplateSource.SPINNAKER_PREFIX
+
 /**
  * Controller for presets
  */
@@ -39,8 +46,17 @@ import org.springframework.web.bind.annotation.RestController
 @RequestMapping('pipelines')
 class PipelineController {
 
-  @Autowired
+  @Autowired(required = false)
+  PipelineTemplateDAO pipelineTemplateDAO = null;
+
   PipelineDAO pipelineDAO
+  ObjectMapper objectMapper;
+
+  @Autowired
+  public PipelineController(PipelineDAO pipelineDAO, ObjectMapper objectMapper) {
+    this.pipelineDAO = pipelineDAO
+    this.objectMapper = objectMapper
+  }
 
   @PreAuthorize("#restricted ? @fiatPermissionEvaluator.storeWholePermission() : true")
   @PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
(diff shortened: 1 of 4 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -82,13 +92,12 @@ class PipelineController {
...
-    if (!pipeline.application || !pipeline.name) {
-      throw new InvalidEntityException("A pipeline requires name and application fields")
-    }
+
+    validatePipeline(pipeline)
+
     pipeline.name = pipeline.getName().trim()
 
     if (!pipeline.id) {
-      checkForDuplicatePipeline(pipeline.getApplication(), pipeline.getName())
```
beckje01: Moved this into the validate call, this will still check for duplicate names in the case of the pipeline id being set which seems like it would be an update case.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -124,21 +133,68 @@ class PipelineController {
     if (pipeline.id != existingPipeline.id) {
       throw new InvalidRequestException("The provided id ${id} doesn't match the pipeline id ${pipeline.id}")
     }
-    pipeline.name = pipeline.getName().trim()
 
-    if (pipelineDAO.getPipelinesByApplication(pipeline.getApplication()).any {
-      it.getName().equalsIgnoreCase(pipeline.getName()) && it.getId() != id }) {
-      throw new DuplicateEntityException("A pipeline with name ${pipeline.getName()} already exists in application ${pipeline.application}")
-    }
+    validatePipeline(pipeline)
```
beckje01: This is after checking that the pipeline ids match which is after a datastore hit I could move this validate to the first thing in the method but that changes error cases in the TCK.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -124,21 +133,68 @@ class PipelineController {
...
+
+    //Check if pipeline type is templated
+    if(pipeline.getType() == "templatedPipeline") {
+      PipelineTemplateDAO templateDAO = getTemplateDAO()
+
+      //Check templated pipelines to ensure template is valid
+      TemplateConfiguration config = objectMapper.convertValue(pipeline.getConfig(), TemplateConfiguration.class);
+
+      //TODO extract magic strings
+      String spinnakerPrefix = "spinnaker://"
```
beckje01: Can I move this to a shared spot so the pipelinetemplate controller can also reference it?

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -124,21 +133,68 @@ class PipelineController {
...
+   */
+  private void validatePipeline(Pipeline pipeline) {
+
+    // Pipelines must has an applicatoin and name
+    if (!pipeline.application || !pipeline.name) {
+      throw new InvalidEntityException("A pipeline requires name and application fields")
+    }
+
+    //Check if pipeline type is templated
+    if(pipeline.getType() == "templatedPipeline") {
```
beckje01: This string should probably be extracted as well.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -39,9 +43,15 @@ import org.springframework.web.bind.annotation.RestController
...
 
+  @Autowired(required = false)
+  PipelineTemplateDAO pipelineTemplateDAO = null;
+
   @Autowired
   PipelineDAO pipelineDAO
 
+  @Autowired
+  ObjectMapper objectMapper;
+
```
robzienert: While you're in here, could you switch the class to use an autowired constructor instead?

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -124,21 +133,68 @@ class PipelineController {
...
-  private void checkForDuplicatePipeline(String application, String name) {
+  /**
+   * Ensure basic validity of the pipeline. Invalid pipelines will raise runtime exceptions.
+   *
+   *
+   * @param pipeline The Pipeline to validate
+   */
+  private void validatePipeline(Pipeline pipeline) {
+
+    // Pipelines must has an applicatoin and name
```
robzienert: `application`

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
robzienert: Add space `if (`.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
robzienert: Const would probably best go on `TemplateConfiguration`?

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -124,21 +133,68 @@ class PipelineController {
...
+      //TODO extract magic strings
+      String spinnakerPrefix = "spinnaker://"
+
+      //With the source check if it starts with "spinnaker://"
+      //Check if template id which is after :// is in the store
+      String source = config.pipeline.template.source
+
+
+      if(source.startsWith(spinnakerPrefix)) {
+        String tempalteId = source.substring(spinnakerPrefix.length())
```
robzienert: `tempalteId`

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
beckje01: Sure

PR #306, opened 2018-04-04
chore(docker): alpine-specific adduser

lwander:


kevinawoo:
added because of https://github.com/spinnaker/spinnaker/issues/2606

PR #311, opened 2018-04-26
fix(oraclebmcs): Throw NotFoundException instead of return null

simonlord:
Fixes the spinnaker/spinnaker#2648 bug where the user sees "Could not create application: Cannot get property 'name' on null object".

Tested locally running deck, redis, gate, orca, front50 - you can now successfully create applications with front50 backed by oracelbmcs storage service.


tomaslin:
LGTM

PR #314, opened 2018-05-11
feat(s3): add regionOverride property

ethanfrogers:
some s3 compatible storage solutions (minio, for example) take the region into account when validating incoming requests but allow you to set non-aws region identifiers. this change allows you to explicitly set the region identifier as an override via `regionOverride`. it also allows you to set the region for configurations using the `endpoint` setting.


ethanfrogers:
@robzienert PTAL
@imosquera FYI

ethanfrogers:
@robzienert can do!

ethanfrogers:
@robzienert documented. i also removed setting the region when it was set to preserve existing functionality. in the end, it wasn't what i was trying to actually fix anyway.

PR #320, opened 2018-06-04
fix(core): Ensure that all pipeline cron triggers have an identifier

ajordens:


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/EnsureCronTriggerHasIdentifierMigration.java
```diff
@@ -0,0 +1,98 @@
...
+      String id = (String) trigger.get("id");
+
+      if ("cron".equalsIgnoreCase(type) && (id == null || id.isEmpty())) {
+        trigger.put("id", UUID.randomUUID().toString());
+      }
+    }
+    dao.update(pipeline.getId(), pipeline);
+
+    log.info(
+      "Added cron trigger identifier (application: {}, pipelineId: {})",
```
dreynaud: isn't that log statement duplicated? Would it make sense to also log the cron id?

ajordens: Was just following a pattern from another migrator where we were logging before and after ... 

Good call on the cron ids ... I've tweaked to include the equivalent of `triggers: ${pipeline.get("triggers")}` ... so we'll get the before/after trigger status.

PR #322, opened 2018-06-05
feat(entitytags): Support for bulk delete of entity tags

ajordens:
Ultimately delegates to the `deleteObjects` available for S3, other
providers will continue to use their respective per-object delete
functions.

```
curl -X "POST" "https://localhost:8081/v2/tags/batchDelete" \
     -H 'Content-Type: application/json' \
     -d $'[
  "aws:servergroup:app-v023:111111111111:us-west-2",
  "aws:servergroup:app-v000:111111111111:us-east-1"
]'
'
```

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.java
@@ -27,6 +27,12 @@ public interface ItemDAO<T> {
 
   void bulkImport(Collection<T> items);
 
+  default void bulkDelete(Collection<String> ids) {
+    for (String id : ids) {
+      delete(id);
+    }
+  }
+
   boolean isHealthy();
 
   default long getHealthIntervalMillis() {
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageService.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageService.java
@@ -38,6 +38,12 @@ public interface StorageService {
 
   void deleteObject(ObjectType objectType, String objectKey);
 
+  default void bulkDeleteObjects(ObjectType objectType, Collection<String> objectKeys) {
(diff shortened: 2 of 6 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/EntityTagsController.groovy
```diff
@@ -91,6 +90,15 @@ class EntityTagsController {
     return findAllByIds(tags.findResults { it.id })
   }
 
+  @RequestMapping(value = "/batchDelete", method = RequestMethod.POST)
```
ajordens: `/batchDelete` seems funny given that the below the fold api is `bulkDelete` ... but we already have a `/batchUpdate` endpoint and I figured it was better to adhere to that pattern.

Internal APIs anyhow.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/ItemDAO.java
```diff
@@ -27,6 +27,12 @@
 
   void bulkImport(Collection<T> items);
 
+  default void bulkDelete(Collection<String> ids) {
+    for (String id : ids) {
```
ajordens: Backwards compatible for the Redis and Cassandra-backed DAOs.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageService.java
```diff
@@ -38,6 +38,12 @@
 
   void deleteObject(ObjectType objectType, String objectKey);
 
+  default void bulkDeleteObjects(ObjectType objectType, Collection<String> objectKeys) {
+    for (String objectKey : objectKeys) {
```
ajordens: Backwards compatible for non-S3 storage services.

ajordens:
This new `batchDelete` endpoint will be eventually called by `clouddriver` to remove entity tags that reference non-existent entities.

PR #326, opened 2018-06-11
fix(google): Fix caching of permissions

ezimanyi:
The GCS ApplicationPermissionDAO is using the modification time of
'specification.json' to determine cache expiration but is actually
reading 'permission.json', leading to incorrect caching.


ezimanyi:
Looks like this was introduced in https://github.com/spinnaker/front50/pull/276, where `keyUpdateTime = service.listObjectKeys(objectType);` was changed to `keyUpdateTime = objectKeyLoader.listObjectKeys(objectType);`.  Since `objectKeyLoader` has the incorrect `service` for permission caching, we're actually pulling the update time of the `specification.json` file rather than the `permission.json` file.

PR #329, opened 2018-06-21
fix(service-accounts): Fiat role sync requires a list of roles not users

dibyom:
We should be sending a list of service account roles for Fiat to refresh, not the serviceAccount names.

See https://github.com/spinnaker/fiat/blob/master/fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/RolesController.java#L99

@stewchen @danielpeach


dibyom:
Changed this to perform a full sync on service account creation.

Fixes spinnaker/spinnaker#2948

PR #340, opened 2018-07-20
fix(core): Fix race condition in cache invalidation

ezimanyi:
For GCS storage, we are currently deciding whether the cache is invalid by comparing a timestamp generated in Java (the time of last refresh) against a timestamp generated by GCS (the time of last update). This can lead to issues with fast read-update-read operations when the two systems don't
have perfectly synchronized clocks. Let's keep track of the last timestamp we actually read from cloud storage, and use that to determine whether the cache should be invalidated.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -60,6 +60,7 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
   private final Counter updateCounter;   // Updates discovered during refresh
 
   private final AtomicLong lastRefreshedTime = new AtomicLong();
+  private final AtomicLong lastSeenStorageTime = new AtomicLong();
 
   public StorageServiceSupport(ObjectType objectType,
                                StorageService service,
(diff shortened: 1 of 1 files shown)


ezimanyi:
This should hopefully fix some intermittent failures we're seeing in our integration tests, where updating a pipeline right after a cache refresh is sometimes not invalidating the cache.

ezimanyi:
@ajordens : Discussed this change with @ewiseblatt , who suggested having you take a look.

ewiseblatt:
@ajordens 

What's going on here in this change is that we're distinguishing between the local clock used to track when the cache was updated against the storage system clock used to signal known changes that affect caching (for polling based storage systems). The two clocks are independent of one another so comparisons dont make sense. The local clock is used to determine how long ago you last updated. The storage clock is essentially a generation number that says it changed since you last took notice of it.

The difference comes into play when you have clock drift. In our case we were a bit ahead of the storage clock, so we thought we updated when really we didnt. Our tests pick this up because they are  intentionally non-forgiving where they want to check white-box outcomes. In practice people might not notice because front50 is self correcting in its periodic refresh scans that dont trust the timestamp markers, or they might hit other replicas than the one that made the change. This PR should improve the timeliness of visibility (within the local replica) in certain clock drift situations.

I'd asked EricZ to run this by you earlier before he'd run some experiments to prove that in fact this is what we were seeing. Even so, maintaining the cache and freshness of data served by front50 is critical for system timeliness. I know you switched over to be more event driven so are probably not affected by this issue, but think you still have a better understanding of subtleties that might be going on here and might pick up on something we havent considered.

ajordens:
This makes sense to me 👍

PR #343, opened 2018-08-09
fix(core): Missing object should not trigger Hystrix fallback

ezimanyi:
There are a number of code paths where we check for the existence of an item by calling findById and catching a NotFoundException if thrown. Currently, when an item is not found in storage, the exception
causes us to fall back to the Hystrix fallback function. This is causing legitimate and expected workflows to trigger the Hystrix circuit breaker; we should not thrown an exception within Hystrix for this case.


ezimanyi:
I also removed the logic where we catch any `HystrixRuntimeException` and re-throw if the fallback exception is a `NotFoundException` as this is no longer needed now that we don't throw a `NotFoundException` from the fallback.

ajordens:
LGTM

ezimanyi:
@ajordens : Thanks!  I'll bump up the queue size for front50 in our CI tests, so they can better deal with the burst of requests that happen when the tests start, and will close this PR.

PR #344, opened 2018-08-09
fix(google): Block updates to lastmodified when update is queued

ezimanyi:
Currently if we fail to update the lastmodified file in storage, due to a 429 (Too Many Requests) response, we queue a request to re-try in 1s. While that request is queued, we should short-circuit any
other attempts to update lastmodified.

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
@@ -540,7 +540,7 @@ GcsStorageService implements StorageService {
         .setName(timestamp_path)
         .setUpdated(new DateTime(System.currentTimeMillis()));
       try {
-        synchronized (this) {
+        synchronized (updateLock(daoTypeName)) {
           // Release the update lock *before* actually updating lastModified as any thread observing
           // the lock as set must know that the last modified time will be updated *after* it observed
           // the lock
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
@@ -16,54 +16,52 @@
 
 package com.netflix.spinnaker.front50.model;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableMap;
-import com.netflix.spinnaker.front50.exception.NotFoundException;
+import static net.logstash.logback.argument.StructuredArguments.value;
 
-import com.google.api.services.storage.Storage;
-import com.google.api.services.storage.StorageScopes;
-import com.google.api.services.storage.model.*;
-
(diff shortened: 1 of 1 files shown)


ezimanyi:
A few comments on the changes here:
* The main goal here was (as noted in the description above) to avoid continuously trying to update the last modified file after we get a 429 (Too many Requests).  While we were catching the 429 and scheduling a re-try after 1s, we were still letting other threads keep trying to update last modified---resulting a even more 429 errors.  Now, we put up a flag to block any other updates to last modified while an update is scheduled---any updates that would have happened during this time can just piggy-back on the already-scheduled one.
* I removed the logic to cancel deferred updates in `scheduleWriteLastModified` as it didn't seem necessary anymore now that there's locking to limit how many concurrent updates are happening.  (There is still a small chance that the deferred update is not needed by the time it runs, but as the only harm is a cache invalidation, I don't think that the cost/complexity here anymore.)
* I updated the retry window to 2s to make it more likely that we won't hit successive 429 errors.  (Given that we're doing an asynchronous update, waiting 2s at the expense of a higher chance of success seems like a reasonable tradeoff, though I suppose this is debatable.)

ezimanyi:
I load tested this, and was able to write 2000 applications in 46 seconds (a rate of 43/second) with no user-facing errors.  Looking in the logs, there were only 28 instances of a "429: Too Many Requests" that had to be handled and re-tried.

Before this change, much lower request rates were triggering so many 429 errors that they started turning into 500's and 503's, and throwing errors back to the caller.

While we should still think about moving away from using the single last-modified file in GCS (perhaps to PubSub notifications) as it represents a scalability bottleneck, I think this change should give us a fair amount of breathing room in the current architecture.

front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -512,33 +512,33 @@ private AtomicBoolean scheduledUpdateLock(String daoTypeName) {
...
-        }
-      };
-      when.setSeconds(when.getSeconds() + 2);
       log.info("Scheduling deferred update {} timestamp.", daoTypeName);
       taskScheduler.schedule(task, when);
     }
   }
 
   private void writeLastModified(String daoTypeName) {
+    // We'll just touch the file since the StorageObject manages a timestamp.
```
ewiseblatt: I dont follow why you moved this outside the guard.

ezimanyi: This was to reduce the amount of code running between acquiring the lock and releasing it, as we'd end up in a bad state if an exception was thrown between acquiring and releasing the lock.  I could have wrapped this in a `try` and released the lock in a `catch` but it seemed cleaner to just move it before acquiring the lock at all...but open to doing the `catch` if you think that's cleaner.

ewiseblatt: you can leave it as is.

PR #345, opened 2018-08-22
feat(google): Add GCS statusCodes to metrics

ewiseblatt:
@ezimanyi

### DIFF
--- a/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
+++ b/front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
@@ -35,6 +35,7 @@ import com.google.api.services.storage.model.Objects;
 import com.google.api.services.storage.model.StorageObject;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.netflix.spectator.api.Clock;
 import com.netflix.spectator.api.Id;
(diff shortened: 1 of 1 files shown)


front50-gcs/src/main/java/com/netflix/spinnaker/front50/model/GcsStorageService.java
```diff
@@ -228,27 +231,40 @@ public GcsStorageService(String bucketName,
...
+       throw e;
      } catch (IOException ioex) {
-         throw ioex;
+       throw ioex;
      } catch (Exception ex) {
-         throw new IllegalStateException(ex);
+       throw new IllegalStateException(ex);
+     } finally {
+       long nanos = clock.monotonicTime() - startTime;
+       String status = Integer.toString(statusCode).charAt(0) + "xx";
```
ezimanyi: In the `deserialize` function, we replace `status` with `200` if it's `<0` (ie, no request was made).  Should we do the same here?

ewiseblatt: The 200 in deserialize is working around in a bug where  a request *is* made and -1 is being returned by the object for some reason. I threw a 200 in there for speculation so it's easier to deal with (I was going to add something like 299 so it would be 2xx but clearly weird but decided against that). The default -1 value would indicate an exception thrown for reasons other than a bad HttpStatus.

PR #346, opened 2018-08-23
 fix(core): Warn when an object's key does not match its id

ezimanyi:
If there are objects in the bucket that front50 is polling that weren't added by front50, some buggy behavior can occur:
* Specific to GCS, objects that match `(.*)[^/]specification.json` are returned by the listing operation, but the actual fetch operation looks for `$1/specification.json` causing a 404.
* For all back-ends, objects whose key does not match do not match their id are cached every cycle.  (Aside from this caching every cycle, looking up the objects by id does not work because the lookup logic assumes the key matches the id.)

The above situations should not occur for any object created by Spinnaker, but let's make the logic robust and/or warn if there are unexpected objects in the bucket.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -330,20 +330,31 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
             .from(ids)
             .flatMap(entry -> {
                   try {
-                    T object = (T) service.loadObject(objectType, entry.getKey());
+                    String key = entry.getKey();
+                    T object = (T) service.loadObject(objectType, key);
 
-                    Long expectedLastModifiedTime = keyUpdateTime.get(entry.getKey());
+                    Long expectedLastModifiedTime = keyUpdateTime.get(key);
                     Long currentLastModifiedTime = object.getLastModified();
 
                     if (expectedLastModifiedTime != null && currentLastModifiedTime != null) {
                       if (currentLastModifiedTime < expectedLastModifiedTime) {
                         log.warn(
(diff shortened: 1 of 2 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -330,22 +330,33 @@ private String buildObjectKey(String id) {
...
                     }
 
+                    if (!key.equals(buildObjectKey(object))) {
+                      log.warn(
+                        "Found an item whose key did not match its expected key.  (key: {}, expected key: {})",
+                        key,
+                        buildObjectKey(object)
+                      );
+                      // Should return Observable.empty() to skip caching, but will wait until the
+                      // logging has been present for a release.
```
jtk54: Any particular reason we are doing this?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -330,22 +330,33 @@ private String buildObjectKey(String id) {
...
-                          entry.getKey(),
+                          key,
                           new Date(currentLastModifiedTime),
                           new Date(expectedLastModifiedTime)
                         );
                       }
                     }
 
+                    if (!key.equals(buildObjectKey(object))) {
+                      log.warn(
```
ewiseblatt: Could the warning be <type> "<key>" has non-matching id "<id>" where:
   <type> is the type of item, 
   <key> is the buildObjectKey
   <id> is key
I'm using "id" here because that's in the user vocabulary. Key here is the cache key, which I understand to be the item's id.

Also, add a spectator counter around this with the item type label so it can be monitored.

PR #354, opened 2018-09-25
feat(s3): adds ability to specify s3 path style, defaults to true

kevinawoo:


kevinawoo:
closing because this might not be needed right now.

PR #355, opened 2018-09-26
refactor(gcs): Update doRetry and bump clouddriver version

dibyom:
Removes the indirect dependency on clouddriver-core

Related to https://github.com/spinnaker/clouddriver/pull/2981

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ allprojects { project ->
   group = "com.netflix.spinnaker.front50"
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.0.13'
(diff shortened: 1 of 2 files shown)


ezimanyi:
@dibyom : I'm getting errors starting front50 after this change.
```org.springframework.beans.factory.BeanDefinitionStoreException: Failed to parse configuration class [com.netflix.spinnaker.front50.Main]; nested exception is org.springframework.context.annotation.ConflictingBeanDefinitionException: Annotation-specified bean name 'retrofitConfig' for bean class [com.netflix.spinnaker.config.RetrofitConfig] conflicts with existing, non-compatible bean definition of same name and class [com.netflix.spinnaker.front50.config.RetrofitConfig]```

It's not obvious to me exactly what changed here, but I can reproducibly start front50 when I sync before this change and then can't when I sync after this change.  Mind taking a look?  Thanks!

PR #364, opened 2018-10-04
fix(s3): Add control over nb of objects retrieved per batch (maxKeys)

ncknt:
Some S3 compatible stores don't support 10000 objects retrieved per object. This PR adds a spinnaker.s3.maxKeys setting to let user override the S3 listObject request. It defaults to 10000 as currently.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.front50.config;
 
+import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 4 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
```diff
@@ -37,6 +43,10 @@ public void setRootFolder(String rootFolder) {
     this.rootFolder = rootFolder;
   }
 
+  public Integer getMaxKeys() { return maxKeys; }
```
lwander: We generally add newlines even to 1-liners

ncknt: Oops, was collapsed in the IDE. I'll update the PR.

PR #367, opened 2018-10-04
fix(s3): Add control over nb of objects retrieved per batch (maxKeys)

ncknt:
This is the same PR as #364 for release 1.10 branch.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.front50.config;
 
+import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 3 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Properties.java
```diff
@@ -29,6 +30,11 @@
   @NestedConfigurationProperty
   S3EventingProperties eventing = new S3EventingProperties();
 
+  // Front50 retrieves objects in batches of this size. Some S3 compatible store enforce a maximum number of keys
+  @Value("${spinnaker.s3.maxKeys:10000}")
```
ajordens: Don't think you should need this `@Value` ... you're already in a `@ConfigurationProperties` class rooted at `spinnaker.s3`.

`Integer maxKeys = 10000`

ncknt:
Thanks. This is for a minio like implementation - not for S3 directly. We get a 400 when `max-keys` is too big.

PR #370, opened 2018-10-05
chore(retrofit): remove unused RetrofitConfig

dreynaud:


dreynaud:
At least I think it's unused, seems to work without it. The commit that introduced it mentioned that fiat needed it, but I currently see `FiatAuthenticationConfig` instantiate its own `OkClient`. Does this look reasonable @ttomsu ?

PR #371, opened 2018-10-06
fix(web): Fix startup failure due to graphql

ezimanyi:
Fetching the schema resource is failing when it's in a jar file as it can't be read as a File. Switch to getting the schema as an InputStream.


ezimanyi:
@robzienert : Was seeing failures starting front50 in a container:
```
java.io.FileNotFoundException: file:/opt/front50/lib/front50-web-1.133.0-SNAPSHOT.jar!/graphql/pipelineConfig.graphqls (No such file or directory)
```
Moved this to get the resource as a stream which should work when it's in a jar.

robzienert:
Argh. Thanks. :(

PR #400, opened 2018-12-04
fix(serviceAccounts): Sync new service accounts with Fiat explicitly

amanya:
We use managed service accounts and immediately after we create them, we need to save the pipeline that uses it. This causes that sometimes the Fiat cache is not fresh enough to know about the new service account and we get 404 in Orca while creating the pipelines with the new roles.
We found that using `loginWithRoles` instead of `sync`, makes Fiat aware of the new service account.


amanya:
@ttomsu PTAL

ttomsu:
LGTM

amanya:
@dibyom Yes, that is definitively the cause, we created an endpoint in fiat to clear that cache and then our workflow worked well.
I made the changes you asked for.

PR #401, opened 2018-12-05
fix(storage/oracle): add private key passphrase

guoyongzhang:
Cherry-pick the fix (#379) into Spinnaker 1.10 release so that it would fix the bucket creation problem in oracle storage.


ttomsu:
I think @duftler is release manager. Matt, note that this is to be merged into 1.10

PR #402, opened 2018-12-05
fix(security): Fix CVE-2017-5929 security issue threat level 9

jingsfdc:
https://nvd.nist.gov/vuln/detail/CVE-2017-5929

QOS.ch Logback before 1.2.0 has a serialization vulnerability affecting the SocketServer and ServerSocketReceiver components.

The RemoteStreamAppenderClient class in logback-classic and the SocketNode classes in logback-classic and logback-access allow data to be deserialized over a Java Socket, via an ObjectInputStream, without validating the data beforehand. When data is received from the Socket, to be logged, it is deserialized into Java objects. An attacker can exploit this vulnerability by sending malicious, serialized Java objects over the connection to the Socket, which may result in execution of arbitrary code when those objects are deserialized.Note that although logback-core is implicated by the Logback project here, the Sonatype Security Research team discovered that the vulnerability is actually present in the logback-classic and logback-access components.Advisory Deviation Notice: The Sonatype Security Research team discovered that this vulnerability was first introduced in version 0.3.0 of logback-classic and version 0.6.0 of logback-access, as opposed to being present in all versions prior to 1.2.0, as stated in the advisory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014 Netflix, Inc.
+ * Copyright (c) 2018, salesforce.com, inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ttomsu:
This should probably be fixed here: https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.template#L44

jingsfdc:
@ttomsu Thanks, will close all PRs and use that one!

PR #403, opened 2018-12-06
feat(app/permission): endpoint for single app permissions

gardleopard:
This endpoint is a building block for making it possible to push new
applications into the fiat application cache when they are created.

Currently FIAT updates its application cache every 20th second and this
will help us being able to trigger updates to the cache when an
application is created.

It will follow a PR to fiat that leverages this endpoint. When both of
these are merged there will come a new front50 PR that triggers the fiat
endpoint for adding a new app so that applications will be available to
users immediately after an application is created

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
@@ -72,6 +72,11 @@ public class PermissionsController {
     return actualPermissions.values()
   }
 
+  @RequestMapping(method = RequestMethod.GET, value = "/applications/{appName:.+}")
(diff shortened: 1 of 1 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
```diff
@@ -72,6 +72,12 @@ public class PermissionsController {
     return actualPermissions.values()
   }
 
+  @ApiOperation(value = "", notes = "Get specific application permissions. Internal use only.")
```
ttomsu: I think technically all Front50 APIs are "internal use only" so I would just remove that.

ttomsu:
@dibyom FYI

PR #439, opened 2019-01-30
fix(test/redis): cloudProviders is promoted to its own field now

jervi:
Broke in https://github.com/spinnaker/front50/pull/427

FYI @ezimanyi


ezimanyi:
@jervi : So it was just the test that was broken?  (I'm surprised the Travis build didn't catch this.)

jervi:
Yes, just the test, as far as I can tell

PR #448, opened 2019-02-11
feat(core): add endpoint for reordering pipelines

maggieneterval:
Adds `/reorder` endpoint for pipelines and pipeline strategies, so that these resources can be re-indexed with one request from Deck rather than one request per pipeline.

Corresponding PRs:
Deck: [spinnaker/deck#6519](https://github.com/spinnaker/deck/pull/6519)
Gate: [spinnaker/gate#723](https://github.com/spinnaker/gate/pull/723)
Orca: https://github.com/spinnaker/orca/pull/2663

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.groovy
@@ -111,4 +111,14 @@ class Pipeline extends HashMap<String, Object> implements Timestamped {
   void setSchema(String schema) {
     super.put("schema", schema)
   }
+
+  @JsonIgnore
+  Integer getIndex() {
+    return (Integer) super.get("index")
+  }
+
+  @JsonIgnore
+  void setIndex(Integer index) {
+    super.put("index", index)
+  }
 }
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ReorderPipelinesController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ReorderPipelinesController.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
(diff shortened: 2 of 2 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -117,6 +117,19 @@ class PipelineController {
     pipelineDAO.bulkImport(pipelines)
   }
 
+  @PreAuthorize("@fiatPermissionEvaluator.storeWholePermission() and hasPermission(#application, 'APPLICATION', 'WRITE')")
```
maggieneterval: @ajordens Wondering what you think the best permission check would be for re-ordering?

ajordens: The tricky thing is that it really needs to be checked on each of the pipeline ids and not the application that was supplied.

Might require explicitly using the `FiatPermissionEvaluator` after each pipeline is loaded to verify that the user has WRITE on the associated application.

(or you need to check that each pipeline is actually associated with the supplied application)

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -117,6 +117,19 @@ class PipelineController {
...
+  @RequestMapping(value = '{application}/reorder', method = RequestMethod.POST)
```
ajordens: I might suggest a new controller for these new operations and an endpoint along the lines of:

```
/actions/pipelines/reorder
```

or

```
/actions/reorderPipelines
```

Per the above comment, the payload could be changed slightly to pass in the application in the body.

maggieneterval:
Thanks so much for your help on this @ajordens and @anotherchrisberry -- should be ready for a second look. Adding @dibyom to double-check the Fiat logic as well.

PR #451, opened 2019-02-12
feat(MPTv2): Adds list by decorated id for artifact support.

jtk54:


ajordens:
FWIW I think `/listIds` is a bit of an odd looking endpoint and collides with the general `/{id}`.

Wherever possible, we should be trying to not repeat the sins of our past.

For example, in this case I think there's an opportunity to just add another parameter to `list()` and return values of the form:
```
[
  {id: "id1"},
  {id: "id2"}
]
```

rather than:
```
[
  "id1",
  "id2"
]
```

Thoughts?

ajordens:
(late feedback -- I missed the original notification that this came through()

jtk54:
I think so as well in hindsight. Currently this is a useful debugging tool but I agree this could be merged into the existing endpoint to make it a bit more reasonable. We're still mid-execution with a lot of this stuff so we can fix it up as we go.

ajordens:
Thanks, appreciate it getting tidied up at some point.

PR #455, opened 2019-02-14
feat(delivery): managed delivery controller and storage

emjburns:
Replacing the old "intent" things with a new `Delivery` based api and model. Any feedback on naming or API structure would be great. 

Per @ajordens suggestion I've slipped a lot of the apis under the `/applications` path.

This is a first stab at defining a managed delivery config. 
Assumptions:
- each application can have more than one config
- you can CRUD them
- you need permissions to do things

Please take a look at the objects and tell me what you think. Most things are defined as maps now.

An example of a delivery config:
```
{
	"application": "emburnstest",
	"lastModifiedBy": "emburns@netflix.com",
	"deliveryArtifacts": [
		{
			"packageName": "spin-package",
			"packageType": "deb",
			"bakeConfig": {
				"baseOs": "xenial",
				"baseLevel": "release",
				"vmType": "hvm",
				"storeType": "ebs"
			},
			"environments": [{
				"name": "test",
				"deployWith": {
					"pipelineId": "my-pipeline-id",
					"pipelineConfig": {}
				},
				"constraints": [
					"pls deploy all the time"
				],
				"order": 1
			}]
		}
	],
	"deliveryEnvironments": [
		{
			"name": "test",
			"infrastructure": [
				{
					"type": "cluster"
				}
			]
		}
	]
}
```


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/DeliveryEnvironment.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.List;
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DeliveryEnvironment {
+  String name;
+  List<Map<String,Object>> infrastructure; //todo eb: let's type this.
```
robfletcher: We should maybe type it as we go. Add definitions here as we support them in Keel. I'm hesitant to suggest sharing a definition but maybe that would be appropriate for once.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/EnvironmentMapping.java
```diff
@@ -0,0 +1,67 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.List;
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class EnvironmentMapping {
+  private String name;
+  private DeployWith deployWith;
+  private List<String> constraints;
```
robfletcher: Does this need to be something more complex than a string? e.g. for the suggested "has to have been in test for > 30 minutes" constraint you might want to make that a generic rule with "test" and "30 minutes" being parameters.

emjburns:
I'm going to follow Cam's suggestion and take a lot of the typing out of here. We will type in Keel.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/EnvironmentMapping.java
emjburns: Yes, probably. Put in a string as a placeholder

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/EnvironmentMapping.java
```diff
@@ -0,0 +1,67 @@
...
+  public int getOrder() {
+    return order;
+  }
+
+  public void setOrder(int order) {
+    this.order = order;
+  }
+
+  public static class DeployWith {
+    private String pipelineId;
```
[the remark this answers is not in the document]
emjburns: I think both, you make a good point that we should support easy options even if we're implementing a pipeline template for people and using that

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/Delivery.java
```diff
@@ -0,0 +1,81 @@
...
+import com.netflix.spinnaker.front50.model.Timestamped;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Delivery implements Timestamped {
+  private String id;
+  private String application;
+  private Long updateTs;
+  private Long createTs;
```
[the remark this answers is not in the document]
emjburns: `Timestamped` is a Long for these data types, I am keeping in the trend so I don't have to do a bunch of conversions

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/Delivery.java
```diff
@@ -0,0 +1,94 @@
...
+  public String getLastModifiedBy() {
+    return lastModifiedBy;
+  }
+
+  @Override
+  public void setLastModifiedBy(String lastModifiedBy) {
+    this.lastModifiedBy = lastModifiedBy;
+  }
+
+  public List<Map<String,Object>> getArtifacts() {
```
cfieber: having just been waist deep in the orca artifacts code I wonder if we should use a different name here (`DeliveryArtifacts` or something?? I hate it but I don't know) just to differentiate between Artifacts in triggers/stages, ExpectedArtifacts, BoundArtifacts, etc, etc

emjburns: Changing to `deliveryArtifacts` because I also hate confusion.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/delivery/Delivery.java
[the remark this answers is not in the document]
ajordens: Ya, I'd keep it consistent for now.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/DeliveryController.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.security.access.prepost.PostAuthorize;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.Collection;
+
+@Slf4j
+@RestController
+@RequestMapping("")
```
ajordens: I don't think you need the empty `@RequestMapping("")`?

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/DeliveryController.java
```diff
@@ -0,0 +1,77 @@
...
+@ConditionalOnExpression("${spinnaker.delivery.enabled:false}")
+public class DeliveryController {
+
+  @Autowired
+  DeliveryRepository deliveryRepository;
```
ajordens: I'd suggest constructor injection as the better practice for new files (and a `private final DeliveryRepository deliveryRepository`).

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/DeliveryController.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @ApiOperation(value = "", notes = "Get all managed delivery configs")
+  @RequestMapping(method = RequestMethod.GET, value = "/delivery")
```
ajordens: Endpoint convention is plural in front50?

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/DeliveryController.java
```diff
@@ -0,0 +1,77 @@
...
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
+  @ApiOperation(value = "", notes = "Get the managed delivery config for an application")
+  @RequestMapping(method = RequestMethod.GET, value = "/applications/{application}/delivery")
+  Collection<Delivery> getConfigByAppName(@PathVariable String application) {
+    return deliveryRepository.getConfigsByApplication(application);
+  }
+
+  @PostAuthorize("hasPermission(returnObject.application, 'APPLICATION', 'READ')")
+  @ApiOperation(value = "", notes = "Get a managed delivery config by id")
+  @RequestMapping(method = RequestMethod.GET, value = "delivery/id/{id}")
```
ajordens: I'd suggest dropping the `/id` and just do `/deliveries/{id}`

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/DeliveryController.java
```diff
@@ -0,0 +1,77 @@
...
+    if (!config.getApplication().equals(application)) {
+      throw new InvalidRequestException("Application in url (" + application + ") and " +
+        "application in config (" + config.getApplication() + ") don't match." );
+    }
+    return deliveryRepository.upsertConfig(config);
+  }
+
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'WRITE')")
+  @ApiOperation(value = "", notes = "Delete a managed delivery config")
+  @RequestMapping(method = RequestMethod.DELETE, value = "/applications/{application}/delivery/id/{id}")
```
ajordens: ditto for removing the `/id`

PR #458, opened 2019-02-15
feat(MPTv2): Update v2 model to use Artifact for template.

jtk54:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/V2TemplateConfiguration.java
```diff
@@ -50,12 +50,11 @@ public void setName(String name) {
...
 
 
-  public TemplateSource getTemplate() {
+  public Artifact getTemplate() {
     return template;
   }
 
-  public void setTemplate(
-    TemplateSource template) {
+  public void setTemplate(Artifact template) {
```
sbwsg: general question about this: doesn't lombok provide these kinds of methods with the `@Data` annotation?

jtk54: Yep. Done.

PR #463, opened 2019-02-20
test(redis): Use embedded redis for unit tests

ezimanyi:
The unit tests for front50-redis depend on there being an existing redis listening on localhost:6379.  If so, they send data to that redis (which may not be intended to receive test data); if not, the tests are silently skipped. Instead, set up a test embedded redis and use that.

This also moves the unhealthy redis tests into their own class, which autowires an unhealthy RedisConnectionFactory. Before this, the tests were mutating the autowired bean, causing issues with test execution order.


ezimanyi:
This should prevent what happened here (#453) where a commit broke these tests because Travis wasn't running them, and they only failed locally for some people (who had a redis running already).

PR #465, opened 2019-02-21
fix(redis): Fix embedded redis leak

ezimanyi:
The refactor of the tests added a redis leak; destroy the bean when it goes out of scope so we don't have a redis outlive the test.


ajordens:
LGTM

PR #466, opened 2019-02-22
fix(authz): Invalidate local Fiat cache on service acct creation.

dibyom:
To ensure that Fiat immediately resolves new permissions involving the newly created service account, we explicitly call Fiat's sync API. We also need to invalidate Front50's in memory cache of fiat permissions. Otherwise, Front50 will continue to use the stale cache to make authorization decisions. 
e.g. if we create a service account and then immediately create a pipeline with this service account, the pipeline save will fail since the current user will not have access to the service account.

Fixes https://github.com/spinnaker/spinnaker/issues/3966


dibyom:
@spinnakerbot cherry-pick 1.12

dibyom:
@spinnakerbot cherry-pick 1.12

PR #476, opened 2019-03-06
feat(authz): Migrate manual service users to managed service users

jervi:
This migration will migrate pipelines that have manual service users to use the new permissions format. It will only run on pipelines where `roles` are not present and `runAsUser` is set to a non-managed service account.
The migration job will automatically create the new managed service users, and the new service user will get the same permissions as the manual service user that it replaces.

The migration must be manually enabled by setting the property `migrations.migrateToManagedServiceAccounts` to `true`.

*Note:* If a pipeline has multiple triggers with _different_ `runAsUser`'s set, the new managed service user will get all of the roles of the different manual service users (as you can only have one managed service user per pipeline). This can potentially remove some users ability to edit or execute affected pipelines. However, I do not think this is a very common scenario.

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigration.java
@@ -93,7 +93,9 @@ public class RunAsUserToPermissionsMigration implements Migration {
       if (runAsUser != null && !runAsUser.endsWith(SERVICE_ACCOUNT_SUFFIX)) {
         ServiceAccount manualServiceAccount = serviceAccounts.get(runAsUser);
         if (manualServiceAccount != null && !manualServiceAccount.getMemberOf().isEmpty()) {
-          List<String> oldRoles = manualServiceAccount.getMemberOf();
+          List<String> oldRoles = manualServiceAccount.getMemberOf().stream()
+              .map(String::toLowerCase) // Because roles in Spinnaker are always lowercase
+              .collect(Collectors.toList());
           newRoles.addAll(oldRoles);
         }
       }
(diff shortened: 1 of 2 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigration.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @Autowired
+  public RunAsUserToPermissionsMigration(PipelineDAO pipelineDAO, ServiceAccountDAO serviceAccountDAO) {
+    this.pipelineDAO = pipelineDAO;
+    this.serviceAccountDAO = serviceAccountDAO;
+  }
+
+  @Override
+  public boolean isValid() {
+    return LocalDate.now(clock).isBefore(VALID_UNTIL);
```
jervi: I'm not sure any time window is really needed for this migration. People will probably be migrating to Pipeline Permissions for a longer time than this, given that it is still an opt-in feature.

dibyom: Agreed. I think this is fine. We can increase this later if needed.

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigration.java
```diff
@@ -0,0 +1,137 @@
...
+import static net.logstash.logback.argument.StructuredArguments.value;
+
+@Component
+@ConditionalOnProperty("migrations.migrateToManagedServiceAccounts")
+public class RunAsUserToPermissionsMigration implements Migration {
+
+  // Only valid until April 1, 2020
+  private static final LocalDate VALID_UNTIL = LocalDate.of(2020, Month.APRIL, 1);
+
+  private static final Logger LOG = LoggerFactory.getLogger(RunAsUserToPermissionsMigration.class);
```
dibyom: Minor style nit: lowercase LOG or use the @Slf4j annotation instead

jervi: Fixed

PR #479, opened 2019-03-08
feat(validation): `PipelineValidator` extension point

ajordens:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/GenericValidationErrors.groovy
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package com.netflix.spinnaker.front50.validator
+
+import org.springframework.validation.AbstractErrors
+import org.springframework.validation.Errors
+import org.springframework.validation.FieldError
+import org.springframework.validation.ObjectError;
+
+class GenericValidationErrors extends AbstractErrors {
```
ajordens: This is basically a copy of what we've got in `ApplicationValidationErrors` ... the two can likely be merged in a subsequent PR.

ajordens: Not sure it's really all that valuable to use Spring `Errors` but we've got this pattern in a few spots already.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -195,6 +203,17 @@ class PipelineController {
...
 
     checkForDuplicatePipeline(pipeline.getApplication(), pipeline.getName().trim(), pipeline.getId())
+
+    def errors = new GenericValidationErrors(pipeline)
+    pipelineValidators.each {
+      it.validate(pipeline, errors)
+    }
+
+    if (errors.hasErrors()) {
+      throw new ValidationException(
```
ajordens: There's built-in support already for `ValidationException` in `kork`.

Applications are doing validations with an Application-specific exception, opt'd not to follow that pattern. We can clean that usage up subsequently as well.

ajordens:
```
{
    "timestamp": 1552003316289,
    "status": 400,
    "error": "Bad Request",
    "exception": "com.netflix.spinnaker.kork.web.exceptions.ValidationException",
    "message": "No message available",
    "errors": ["your error message here"]
}
```

Serialized response fits the pattern we've got elsewhere.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/GenericValidationErrors.groovy
robzienert: I think I'd prefer to use `Errors` over something else, personally. Converge on a standard, and one that we don't have to write 💯

front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/PipelineValidator.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.front50.validator;
+
+import com.netflix.spinnaker.front50.model.pipeline.Pipeline;
+import org.springframework.validation.Errors;
+
+public interface PipelineValidator {
```
robzienert: I'd like to start seeing javadocs on all interfaces (class- & method-level) we introduce. This one is pretty explanatory, but it'll be a good habit to get into regardless.

ajordens: :thumbs_up: 

I like and can get behind that.

PR #488, opened 2019-03-14
fix(pipelines): More robust handling of missing cron trigger id's

jervi:
Needed because we sometimes generate the pipeline id outside Front50 (e.g. https://github.com/spinnaker/orca/pull/2754)

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
@@ -108,15 +108,19 @@ class PipelineController {
     pipeline.name = pipeline.getName().trim()
     pipeline = ensureCronTriggersHaveIdentifier(pipeline)
 
-    if (!pipeline.id) {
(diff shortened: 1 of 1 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
```diff
@@ -108,17 +108,21 @@ class PipelineController {
     pipeline.name = pipeline.getName().trim()
     pipeline = ensureCronTriggersHaveIdentifier(pipeline)
 
-    if (!pipeline.id) {
+    if (!pipeline.id || extractCronTriggers(pipeline).any { !it.id }) {
       // ensure that cron triggers are assigned a unique identifier for new pipelines
-      def triggers = (pipeline.triggers ?: []) as List<Map>
-      triggers.findAll { it.type == "cron" }.each { Map trigger ->
+      extractCronTriggers(pipeline).each { Map trigger ->
```
xavileon: filter those that have an `id` already?

jervi: Will fix

PR #489, opened 2019-03-14
chore(syncroles): Sync only the service account roles

xavileon:
Doing a full sync implies querying all the service accounts back to
front50 when doing the sync. However, we only need to pass the roles of
the service user being created. This helps on sync time when the amount
of service accounts is huge.


xavileon:
@dibyom PTAL, this is ready to be merged now after https://github.com/spinnaker/fiat/pull/356 was merged 🙂

PR #493, opened 2019-03-19
fix(authz): Automatic Service Account migration missing some pipelines

jervi:
If a pipeline has two or more triggers, but not all are updated to automatic service accounts, the pipeline would still be skipped in the migration because `roles` would already be present. This PR fixes so that existing roles will be handled and triggers with manual service users are always migrated.

### DIFF
--- a/front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigrationSpec.groovy
+++ b/front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigrationSpec.groovy
@@ -16,12 +16,16 @@ import java.time.Clock
 import java.time.Instant
 import java.time.ZoneId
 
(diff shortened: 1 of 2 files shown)


front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/RunAsUserToPermissionsMigrationSpec.groovy
```diff
@@ -23,6 +23,10 @@ class RunAsUserToPermissionsMigrationSpec extends Specification {
   @Subject
   def migration = new RunAsUserToPermissionsMigration(pipelineDAO, serviceAccountDAO)
 
+  def setup() {
+    migration.setClock(Clock.fixed(Instant.parse("2019-04-01T10:15:30.00Z"), ZoneId.of("Z")))
```
jervi: Stops the tests from starting to fail in a year or so :)

PR #502, opened 2019-03-28
fix(MPTv2): Avoids looking up v2 MPTs by 'latest'.

jtk54:


jtk54:
@spinnakerbot cherry-pick 1.13

PR #509, opened 2019-04-09
feat(MPTv2): Updates latest tag on v2 MPT write.

jtk54:


jtk54:
Semantics aren't totally correct yet in this PR, needs more work. Closing for now.

PR #510, opened 2019-04-09
feat(MPTv2): Updates latest tag on v2 MPT write.

jtk54:
Also: 

* made some fixes to digest last updated time handling -- previously on each update we were calculating a different hash due to the update time rolling.

* fixed strange error handling in `/create` where we would no-op and succeed when trying to write a duplicate id + tag combo.


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/PipelineTemplate.java
```diff
@@ -89,6 +89,11 @@ public Long getLastModified() {
     return (updateTs != null) ? Long.valueOf(updateTs) : null;
   }
 
+  public Long removeLastModified() {
+    String updateTs = (String) super.remove("updateTs");
```
ajordens: this doesn't seem to be doing what the method name would imply?

jtk54:
The method name is referring to the Map remove() function and the behavior
should be identical (just null safe). I'll drop a comment to clarify.

PR #516, opened 2019-04-19
feat(fiat): Adds migrator to add execute permissions to applications.

dibyom:
Execute is added to all groups that currently have READ or WRITE permissions on an application. This is because currently anyone with READ permission can execute a pipeline.

Checks enforcing that users need EXECUTE to trigger a pipeline is coming in a follow up PR. 

Part of spinnaker/spinnaker#3554

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/ApplicationExecutePermissionMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/ApplicationExecutePermissionMigration.java
@@ -1,0 +1,68 @@
+package com.netflix.spinnaker.front50.migrations;
+
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.front50.model.application.Application;
+import com.netflix.spinnaker.front50.model.application.ApplicationPermissionDAO;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+
+import java.time.Clock;
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.ArrayList;
+
+@Component
+@Slf4j
+public class ApplicationExecutePermissionMigration implements Migration {
+
+  // Only valid until May 1st, 2020
+  private static final LocalDate VALID_UNTIL = LocalDate.of(2020, Month.APRIL, 1);
+
+  private ApplicationPermissionDAO applicationPermissionDAO;
+
+  @Setter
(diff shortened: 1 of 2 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/ApplicationExecutePermissionMigration.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final LocalDate VALID_UNTIL = LocalDate.of(2020, Month.APRIL, 1);
+
+  private ApplicationPermissionDAO applicationPermissionDAO;
+
+  @Setter
+  private Clock clock = Clock.systemDefaultZone();
+
+  @Autowired
+  public ApplicationExecutePermissionMigration(
+      ApplicationPermissionDAO applicationPermissionDAO) {
```
ezimanyi: nit: indentation is a bit off here

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/ApplicationExecutePermissionMigration.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.ArrayList;
+
+@Component
+@Slf4j
+public class ApplicationExecutePermissionMigration implements Migration {
+
+  // Only valid until May 1st, 2020
+  private static final LocalDate VALID_UNTIL = LocalDate.of(2020, Month.APRIL, 1);
```
ezimanyi: There's a mismatch between the comment and the code (April vs. May)

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/ApplicationExecutePermissionMigration.java
```diff
@@ -0,0 +1,68 @@
+package com.netflix.spinnaker.front50.migrations;
```
ezimanyi: Missing header (both files)

ajordens:
What are your thoughts on handing the lack of explicit `EXECUTE` when someone is setting permissions on an application ... the migrator would _eventually_ detect it but not immediately.

ajordens:
Good point about the fallbacks ... 

It seems plausible that we might be able to get away without a migration, as well as configuration point to determine how `EXECUTE` should be interpreted? The later may satisfy a desire of ours to move from `READ` == `EXECUTE` to `WRITE` == `EXECUTE`.

Could be simpler without much of a hit complexity wise in `FiatPermissionEvaluator`?

jtk54:
Code looks good to me, but +1 to ajordens suggestion to explore whether we can get out of jail free without a migrator.

dibyom:
@ezimanyi I like the idea. I'll just add the fallback check in FiatPermissionEvaluator. Defaults to anyone with read has execute but the option to change to write implies execute for those that want it. I don't think there would be any performance hit.

PR #523, opened 2019-05-07
feat(s3): Support for assumeRole

ncknt:
This adds support for assume role to s3. It's using a new `assumeRole` (which could be a role or an ARN) and `assumeRoleAccountId` which is used when `assumeRole` is not an ARN.

This implements what the `--assume-role` is meant for in Halyard (currently doesn't do anything): https://www.spinnaker.io/reference/halyard/commands/#hal-config-storage-s3-edit. We'll add the `--assume-role-account-id` in Halyard as well.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3Config.java
@@ -3,6 +3,7 @@ package com.netflix.spinnaker.front50.config;
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.Protocol;
 import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.regions.Region;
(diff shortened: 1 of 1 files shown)


ncknt:
@ajordens, @emjburns: could you PTAL? Unsure if we prefer to use `NetflixSTSAssumeRoleSessionCredentialsProvider` or the newer `STSAssumeRoleSessionCredentialsProvider.Builder`.

robzienert:
We want to use the `Netflix...` one. Better to be consistent, then if we want to change the behavior of assume role, we only need to do it in one place.

robzienert:
Do we still want to merge this? PR has conflicts.

ncknt:
Let me take a look...

ncknt:
This PR was relying on `clouddriver-aws` which is no longer a dependency since Spring Boot 2 clean up. Re-adding it doesn't seem fun (kork-bom?). I'll check if that's still needed, obviously we've been working around it until now.

ncknt:
It is not currently needed. Closing the PR. I'll open a new one if this comes up again.

PR #527, opened 2019-05-17
fix(core): Switch location to additional-location

ezimanyi:
In boot2, location overrides rather than adds to the default config locations.


ezimanyi:
Caught by our integration tests; similar to what we did in other services, ex spinnaker/echo#536

PR #530, opened 2019-05-21
fix(core): Ensure lastModified and lastModifiedBy is set on item

ajordens:
Currently we're relying on the underlying `StorageService` implementations
to set and track lastModified and lastModifiedBy.

Pulling it up to the `StorageServiceSupport` ensures consistent behavior
across all current and future `StorageService` implementations.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -219,6 +219,8 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
   }
 
   public void update(String id, T item) {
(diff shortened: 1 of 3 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -219,6 +219,8 @@ public T findById(String id) throws NotFoundException {
   }
 
   public void update(String id, T item) {
+    item.setLastModifiedBy(AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"));
+    item.setLastModified(System.currentTimeMillis());
```
ajordens: The object store backed services currently override the `lastModified` at load time but it's confusing if you're ever looking at the raw records outside of `front50`.

PR #553, opened 2019-07-08
feat(sql): Support for a SQL `StorageService` implementation

ajordens:
This PR maintains backwards compatibility and does not modify existing
controllers.

The `CompositeStorageService` allows for dual-read/write between two
storage services (ie. SQL and non-SQL).

The `StorageServiceMigrator` can be used to migrate objects between
two storage services (ie. non-SQL to SQL).

Enabling SQL:

```
sql:
    enabled: true
    connectionPools:
    default:
        jdbcUrl: jdbc:mysql://.....
        user: root
        password: null
        default: true
    migration:
    jdbcUrl: jdbc:mysql://.....
    user: root
    password: null
```

Enabling the `CompositeStorageService`:

```
spinnaker:
    migration:
      compositeStorageService:
        enabled: true
        reads:
          primary: false
          previous: true
```

Enabling the `StorageServiceMigrator`:

```
spinnaker:
    migration:
    enabled: true
    primaryClass: com.netflix.spinnaker.front50.model.SqlStorageService
    previousClass: com.netflix.spinnaker.front50.model.S3StorageService
```


ajordens:
This is a lift of the SQL implementation we've been running `front50` with for the past couple of months.

PR #567, opened 2019-08-12
[DNMY] feat(authz): More robust handling of managed service accounts

jervi:
Reverts #488, as it is no longer needed. Adds two new endpoints to the `/serviceAccounts` endpoint: one to request a single service account and one to invalidate local fiat caches.
Prerequisite for https://github.com/spinnaker/orca/pull/3092


cfieber:
this could probably align with #588 for asserting a CREATE permission on serviceAccount once that one lands

PR #571, opened 2019-08-16
feat(proto): Adding front50 application proto definition

robzienert:
See: https://github.com/spinnaker/spinnaker/issues/4699

An initial pass at defining a proto for applications. Please critique thoroughly.

no idea what i'm doing

### DIFF
--- a/front50-proto/front50-proto.gradle
+++ b/front50-proto/front50-proto.gradle
@@ -1,0 +1,59 @@
+buildscript {
+  repositories {
+    jcenter()
+    maven { url "https://spinnaker.bintray.com/gradle" }
+    maven { url "https://plugins.gradle.org/m2/" }
+  }
+  dependencies {
+    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.8'
+  }
+}
+
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
+
+dependencies {
+  // TODO(rz): Promote these dep versions into kork-dependencies
+  implementation 'com.google.protobuf:protobuf-java:3.0.0'
+  implementation 'com.google.guava:guava:23.5-jre'
+  api 'com.google.api.grpc:grpc-google-common-protos:1.0.5'
+  implementation 'io.grpc:grpc-all:1.8.0'
+}
+
+protobuf {
+  protoc {
+    artifact = 'com.google.protobuf:protoc:3.5.1-1'
+  }
+  plugins {
+    grpc {
+      artifact = "io.grpc:protoc-gen-grpc-java:1.9.0"
+    }
+  }
+
+  generateProtoTasks {
+    ofSourceSet('main').each { task ->
+      task.builtins {
+        java{
+          outputSubDir = 'java'
+        }
+      }
+      task.plugins {
+        grpc {
+          outputSubDir = 'java'
+        }
+      }
+      task.descriptorSetOptions.includeImports = true
+    }
+  }
+  generatedFilesBaseDir = "${projectDir}/build-gen/proto"
+}
+
+idea {
+  module {
+    sourceDirs += file("${projectDir}/build-gen/proto/main/java")
+  }
+}
+
+clean {
+  delete "${projectDir}/build-gen"
+}
--- a/front50-proto/src/main/proto/applications.proto
+++ b/front50-proto/src/main/proto/applications.proto
@@ -1,0 +1,242 @@
+syntax = "proto3";
+
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+package spinnaker.front50.v3;
+
+option java_package = "com.netflix.spinnaker.front50.proto";
+option java_multiple_files = true;
+
+// ApplicationService is the service for interacting with the Spinnaker application domain.
+service ApplicationService {
+  // ListApplications provides a way of searching through Applications..
+  rpc ListApplications (ListApplicationsRequest) returns (ListApplicationsResponse);
+
+  // GetApplication returns a single Application matching the criteria provided.
+  rpc GetApplication (GetApplicationRequest) returns (Application);
+
+  // CreateApplication will create a new Application. If an Application already exists by its name, this operation
+  // will raise a PreconditionFailure.
+  rpc CreateApplication (CreateApplicationRequest) returns (Application);
+
+  // UpdateApplication will update an existing Application. If an Application does not exist at the provided name, this
+  // operation will raise a PreconditionFailure.
+  rpc UpdateApplication (UpdateApplicationRequest) returns (Application);
+
+  // DeleteApplication will delete an Application. This operation is unaffected by prior state.
+  rpc DeleteApplication (DeleteApplicationRequest) returns (google.protobuf.Empty);
+
+  // ListApplicationRevisions returns an Application's revision history.
+  rpc ListApplicationRevisions (ListApplicationRevisionsRequest) returns (ListApplicationRevisionsResponse);
+}
+
+// An Application includes all configuration for a particular service within Spinnaker, and is an aggregate of
+// associated infrastructure collections such as clusters, server groups, firewalls, load balancers, etc.
+//
+// You will typically create a different application for each service, though Spinnaker does not enforce that.
+message Application {
+  // Required. Resource name for the application (e.g. "applications/front50")
+  string name = 1;
+
+  // Required. The official name of the application (e.g. "front50")
+  string title = 2;
+
+  // The description of the application's purpose and function.
+  string description = 3;
+
+  // Required. The email address (typically a distribution list) of the owner to the application.
+  repeated string owner_email = 4;
+
+  // Output only. When the application was first created.
+  google.protobuf.Timestamp create_time = 5;
+
+  // Output only. When the application was last updated.
+  google.protobuf.Timestamp update_time = 6;
+
+  // Output only. The actor which updated the application.
+  string updated_by = 7;
+
+  // A list of cloud providers that the application has configured.
+  //
+  // While not required, defining cloud providers will help inform Spinnaker's behavior for the application.
+  repeated string cloud_providers = 8;
+
+  // A bucket of named, client-typed application attributes.
+  //
+  // This allows extension of the configuration that Front50 provides for an application without modification of its
+  // core data model.
+  map<string, AttributeGroup> attribute_groups = 9;
+
+  // A named collection of instance links for an application.
+  repeated InstanceLinkGroup instance_links = 10;
+
+  // A list of all accounts an application resides in.
+  //
+  // While not required, defining accounts will help inform Spinnaker's behavior for the application.
+  repeated string accounts = 11;
+
+  // A list of application-wide notification configurations.
+  //
+  // Notifications can be triggered in response to a number of events within Spinnaker, including pipeline execution
+  // lifecycle changes.
+  repeated Notification notifications = 12;
+
+  // An arbitrary bucket of labels for the application that can be used by end-users and integrators alike.
+  map<string, string> labels = 13;
+}
+
+// AttributeGroup is a named, client-typed bucket of data used for extending Front50's data model.
+message AttributeGroup {
+  // Resource name for the attribute group (e.g. "attributeGroups/netflixTimelines")
+  string name = 1;
+
+  // The data associated with the attribute group.
+  google.protobuf.Struct data = 2;
+
+  // The required permissions for updating an attribute group.
+  repeated Permission permissions = 3;
+}
+
+// InstanceLinkGroup is a titled collection of InstanceLinks.
+message InstanceLinkGroup {
+  // The title of the group.
+  string title = 1;
+
+  // A collection of instance links.
+  repeated InstanceLink links = 2;
+}
+
+// InstanceLink is a titled URL link template for server group instances.
+message InstanceLink {
+  // The title of an instance link.
+  string title = 1;
+
+  // The link template value.
+  string link = 2;
+}
+
+// Notification is the configuration for application-wide notification recipients.
+message Notification {
+  // The type of notification (e.g. slack, email, webhook...)
+  string notification_type = 1;
+
+  // The type-specific notification recipient (e.g. email address, slack channel(s), webhook endpoint...)
+  string recipient = 2;
+
+  // A collection of event names to notify on.
+  repeated string notifying_events = 3;
+
+  // A bucket of type-specific notification config.
+  google.protobuf.Struct metadata = 6;
+}
+
+// Permission defines an authorization and its allowed prinicpals.
+message Permission {
+  // The privilege that is being granted.
+  Privilege name = 1;
+
+  enum Privilege {
+    UNKNOWN = 0;
+    READ = 1;
+    WRITE = 2;
+    EXECUTE = 3;
+  }
+
+  // A collection of principals who are allowed this permission.
+  repeated string allowed_principals = 2;
+}
+
+// ListApplicationsRequest can be used to list and search applications known to Spinnaker.
+message ListApplicationsRequest {
+  // The number of applications to return in the response.
+  //
+  // Defaults to 100.
+  int32 page_size = 1;
+
+  // The page token to continue pagination from. If left unset, the first `page_size` results will be returned.
+  string page_token = 2;
+
+  // Allows filtering of applications by fields.
+  string filter = 3;
+
+  // Order applications by fields and direction.
+  string order_by = 4;
+
+  // Define partial responses of an application.
+  google.protobuf.FieldMask fields = 5;
+}
+
+// ListApplicationsResponse returns a page of applications matching the provided request.
+message ListApplicationsResponse {
+  // The applications returned.
+  repeated Application applications = 1;
+
+  // The next page token, if any.
+  string next_page_token = 2;
+}
+
+// GetApplicationRequest returns a single application.
+message GetApplicationRequest {
+  // Required. Resource name for the application (e.g. "applications/front50")
+  string name = 1;
+
+  // Define partial responses of an application.
+  google.protobuf.FieldMask fields = 2;
+}
+
+// CreateApplicationRequest will attempt to create an application.
+message CreateApplicationRequest {
+  // Required. The application to be created.
+  Application application = 1;
+
+  // If true, it indicates that the given request should only be validated, not executed.
+  bool validate_only = 2;
+}
+
+// UpdateApplicationRequest will attempt to update an existing application.
+message UpdateApplicationRequest {
+  // Required. Resource name for the application (e.g. "applications/front50")
+  string name = 1;
+
+  // Required. The application payload to update with.
+  Application application = 2;
+
+  // Allows partial update on an application. This mask is relative to the resource, not to the request message.
+  google.protobuf.FieldMask update_mask = 3;
+
+  // If true, it indicates that the given request should only be validated, not executed.
+  bool validate_only = 4;
+}
+
+// DeleteApplicationRequest will delete an application.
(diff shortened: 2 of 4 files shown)


front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  string page_token = 3;
+}
+
+// ListApplicationRevisionsResponse
+message ListApplicationRevisionsResponse {
+  // The revisions of an application.
+  repeated Application revisions = 1;
+
+  // The next page token, if any.
+  string next_page_token = 3;
```
ezimanyi: Should this be `= 2`?

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  repeated string notifying_events = 3;
+
+  // A bucket of type-specific notification config.
+  map<string, string> metadata = 6;
+}
+
+// Permission defines an authorization and its allowed prinicpals.
+message Permission {
+  // The name of the authorization (e.g. WRITE, READ, EXECUTE...)
+  string name = 1;
```
ezimanyi: Would we want this to be an enum?

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  // The number of applications to return in the response.
+  //
+  // Defaults to 100.
+  int32 page_size = 1;
+
+  // The page token to continue pagination from. If left unset, the first `page_size` results will be returned.
+  string page_token = 2;
+
+  // Allows filtering of applications by fields.
+  string filter = 3;
```
ezimanyi: What's an example of what this filter would look like?

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+
+  // Order applications by fields and direction.
+  string order_by = 4;
```
brianwolfe: Should this be more opinionated (enum of valid ordering options?). Leaving this as a string makes forwards/backwards compatibility harder to validate at the proto level.

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  string name = 1;
+
+  // Required. The official name of the application (e.g. "front50")
+  string title = 2;
+
+  // The description of the application's purpose and function.
+  string description = 3;
+
+  // Required. The email address (typically a distribution list) of the owner to the application.
+  string owner_email = 4;
```
brianwolfe: We have some cases of joint ownership, where this becomes `repeated`.

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+// ApplicationService is the service for interacting with the Spinnaker application domain.
+service ApplicationService {
+  // ListApplications provides a way of searching through Applications..
+  rpc ListApplications (ListApplicationsRequest) returns (ListApplicationsResponse);
+
+  // GetApplication returns a single Application matching the criteria provided.
+  rpc GetApplication (GetApplicationRequest) returns (Application);
+
+  // CreateApplication will create a new Application. If an Application already exists by its name, this operation
+  // will fail.
```
brianwolfe: We should probably be specific about the type of [errors](https://www.grpc.io/docs/guides/error/) we should raise. Something like the precondition failure seems appropriate here? https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L93-L115

front50-proto/src/main/proto/applications.proto
brianwolfe: nm, this is directly from https://cloud.google.com/apis/design/design_patterns

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  string notification_type = 1;
+
+  // The type-specific notification recipient (e.g. email address, slack channel(s), webhook endpoint...)
+  string recipient = 2;
+
+  // A collection of event names to notify on.
+  repeated string notifying_events = 3;
+
+  // A bucket of type-specific notification config.
+  map<string, string> metadata = 6;
```
plumpy: This too could maybe be an Any, where each notification type could define their own proto message?

Using Any definitely adds some amount of overhead to the client code, but once you resolve the Any to a specific type of proto message, you get all the benefits of a proto instead of a Map. So it's kinda debatable but I'd lean towards Any personally...

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  // Defaults to 100.
+  int32 page_size = 1;
+
+  // The page token to continue pagination from. If left unset, the first `page_size` results will be returned.
+  string page_token = 2;
+
+  // Allows filtering of applications by fields.
+  string filter = 3;
+
+  // Order applications by fields and direction.
```
plumpy: Are we only sorting by a single field, so it's just SQL-style `fieldname desc`? Anything more complicated probably deserves an `Ordering(string field_name, enum Order order)` message...

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+// CreateApplicationRequest will attempt to create an application.
+message CreateApplicationRequest {
+  // Required. The application to be created.
+  Application application = 1;
+
+  // If true, it indicates that the given request should only be validated, not executed.
+  bool validate_only = 2;
+
+  // A unique string id used for detecting duplicated requests.
+  string request_id = 3;
```
plumpy: How is this going to work? Does the created/updated application contain the `last_request_id` and if we see `request_id==last_request_id` we return success?

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  string page_token = 2;
+
+  // Allows filtering of applications by fields.
+  string filter = 3;
+
+  // Order applications by fields and direction.
+  string order_by = 4;
+
+  // Define partial responses of an application.
+  google.protobuf.FieldMask fields = 5;
```
plumpy: Does anyone care about this? (Non-rhetorical question!) I imagine not, in which case I wouldn't bother implementing. But maybe I'm wrong.

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  string request_id = 5;
+}
+
+// DeleteApplicationRequest will delete an application.
+message DeleteApplicationRequest {
+  // Required. Resource name for the application (e.g. "applications/front50")
+  string name = 1;
+}
+
+// ListApplicationRevisionsRequest will return a list of application history revisions.
```
plumpy: ...ordered from newest to oldest, presumably?

front50-proto/src/main/proto/applications.proto
robzienert: Hmm, I was 99% sure I saw this format defined in the Google Cloud API design doc, but now that I'm searching, I can't seem to find it. :\ From memory, it was something like `myField=foo;myField2=Blah`. Does this ring any bells for you Google folks?

front50-proto/src/main/proto/applications.proto
robzienert: Yes, agree - from what I understand of gRPC (which is admittedly not much yet), these effort formats are not actually defined as part of the response, but instead handled through exceptions. Isn't that correct? I'm operating off of this page: https://cloud.google.com/apis/design/errors

front50-proto/src/main/proto/applications.proto
```diff
@@ -0,0 +1,241 @@
...
+  map<string, string> labels = 13;
+}
+
+// AttributeGroup is a named, client-typed bucket of data used for extending Front50's data model.
+message AttributeGroup {
+  // Resource name for the attribute group (e.g. "attributeGroups/netflixTimelines")
+  string name = 1;
+
+  // The data associated with the attribute group.
+  google.protobuf.Struct data = 2;
```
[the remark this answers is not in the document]
robzienert: You tell me: I'm not clear on when we'd want to use `Any` vs `Struct`? These attribute groups (internally) are 90% of the time a JSON object, which is why I chose `Struct`. If there's an edge case of data (some attribute groups are actually just a single scalar value), should I be more inclined towards `Any`?

front50-proto/src/main/proto/applications.proto
robzienert: Thanks, that makes sense. When using an `Any`, is it common practice to have an associated `kind` field (in this case, `metadataKind`) as well to help resolution of the embedded proto?

front50-proto/src/main/proto/applications.proto
robzienert: My rationale here was that our various services often request applications from front50, but only need 1 or 2 fields from it. Gate and Echo aside, we always request application data on-demand rather than keeping an in-memory near cache.

Specificially for Deck, there's been a lot of internal desire from our frontend engineers to slim up our responses from Gate, which this would also help with.

front50-proto/src/main/proto/applications.proto
plumpy: Most of the compute APIs have a similar field, e.g.:
https://cloud.google.com/compute/docs/reference/rest/v1/regionInstanceGroupManagers/list

But I don't think it's a standard syntax for which we have an open source parser, so you'll probably have to write your own...

front50-proto/src/main/proto/applications.proto
robzienert: I wasn't planning on requiring an optimistic locking model, instead having a client-generated hash that would be used strictly for de-duplication. Certainly open to feedback here, I definitely see the benefits of what you're proposing.

front50-proto/src/main/proto/applications.proto
plumpy: Any includes a field `type_url` to say what protobuf message it contains:
  https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto
So we could set it to things like `spinnaker.io/com.netflix.notifications.SlackMessage` or whatever. I suppose in this case, you could probably even get rid of `notification_type`, since the type of the `Any` message would tell you what kind of notification we have.

But maybe for front50 we don't want to go that route? Maybe that's shoving too much knowledge of the types into front50? Hmm...

front50-proto/src/main/proto/applications.proto
plumpy: Just in case the client accidentally sent the same request twice, basically? But what if those requests got load-balanced to different front50 instances or something... it seems like a hard problem to solve well, so I'd lean in favor of not trying to solve it at all, but maybe I'm just being difficult.

front50-proto/src/main/proto/applications.proto
plumpy: Yeah, okay, that seems totally reasonable then, thanks.

front50-proto/src/main/proto/applications.proto
brianwolfe: I think you are correct (also not a gRPC expert). This isn't explicit in the response object. When documenting the API, we should probably be specific about particular anticipated errors.

front50-proto/src/main/proto/applications.proto
brianwolfe: This is explicitly the form of the Google APIs, as specified here: https://cloud.google.com/apis/design/design_patterns in Sorting Order.

front50-proto/src/main/proto/applications.proto
brianwolfe: https://blog.envoyproxy.io/dynamic-extensibility-and-protocol-buffers-dcd0bf0b8801 has a decent description of any vs. struct. In particular, the following stick out to me:

Any:
> There is no meaningful way to make sense of the embedded proto without its schema, i.e. its protobuf descriptor.
> Messages can be efficiently (de)serialized, without JSON round-tripping.

Struct:
> This is a very flexible type and brings the advantages of dynamic typing to protobuf.
> It’s not possible to statically type check.
> This representation is not wire efficient.
> The official language specific protobuf libraries do not have first-class support for converting between Struct and arbitrary protobuf message types, instead it’s always necessary to round-trip via JSON (de)serialization ops.

Seems like the tradeoff is around how opaque the types should be from Front50's perspective. Should front50 be storing opaque blobs? Or should we require that it is extended to know about additional Message types. Based on the envoy blog post above, seems like it might be case-by-case, depending on if the API is considered an opaque extension point (biased towards a Struct) or a standard component that Front50 may need to interpret in the future (arguing for an Any).

plumpy: Yeah, good summary, thanks. For a higher-level service like Echo that deals more directly with notifications, I definitely would not expect it to be dealing in structs, and Any probably makes a lot more sense. But for a data service like front50, those considerations change lot, and maybe Any imposes too much structure...

front50-proto/src/main/proto/applications.proto
robzienert: I already came to terms with writing a parser. Imagine we applied the same `filter` field & parser across all of our APIs where filtering functionality was required - how would you feel about this?

front50-proto/src/main/proto/applications.proto
robzienert: 💯 on the documentation side of things. Since it's not part of the payload, I'm kinda biased towards documenting in a single location similar to the links we provided each other; rather than mentioning in each proto that X proto can return Y, Z, N... errors. WDYT?

front50-proto/src/main/proto/applications.proto
robzienert: OK, that was helpful! Sounds like Struct is the better fit basically across the board for front50 at this point.

ezimanyi: Given that `front50` is going to be the owner of the data (rather than just a pass-through) I'd probably vote for stronger typing and going with the `Any`.  If it raises the bar to adding a new notification type, I actually think that might be a good thing (as that bar would be raised by forcing more structure and better documentation).

My opinion is probably strongly affected by spending a lot of time wondering "what is in this map" for intra-service functions, but I'd love to try as much as possible to avoid that for things in the API.  That being said, this lack of typing is limited to one small corner of the application object, so maybe it's reasonable here...

front50-proto/src/main/proto/applications.proto
brianwolfe: For this particular comment, I'd prefer moving:

> this operation will fail.

to

> this operation will raise a PreconditionFailure.

brianwolfe: I don't think an exhaustive list is necessary, but precision in particularly common exceptions we anticipate seems worthwhile.

front50-proto/src/main/proto/applications.proto
robzienert: We have some orgs that have a similar setup (like in the case of transitioning ownership in reorgs, etc). I wonder if having something like `repeated Owner(string email, string label, bool primary)` might make more sense? It feels weird to me to have more than 1 owner. Interested to hear your thoughts: In practice, would additional metadata (like a label, primary flag, etc) reduce any confusion about ownership / who someone should contact for a particular context?

front50-proto/src/main/proto/applications.proto
brianwolfe: I don't feel like I have enough experience with systems like this to judge where we'll hit future hurdles. How responsible is Front50 for ensuring its own data integrity? At Airbnb, we are currently extending Gate, Echo, Kayenta, Orca, and Clouddriver. We are not currently extending Front50. This is nice from an overhead perspective, but it also means that Front50 doesn't really have the knowledge to:

1. enforce any correctness when another service stores a notification type (if two systems have different ideas about the schema they want, we lose the value of typing here).
2. migrate data from one schema to another when performing major version (non-backwards compatible) upgrades.

Maybe we should be aiming to make it easy to include new Message types as plugins for each service instead of optimizing for maximal flexibility at runtime? I feel like I'd rather that flow be super easy than have to deal with all the pain around adding manual schema-checking logic on the echo side for each notification type.

robzienert: I don't disagree, but for conversation sake: I'd like to reduce use cases where we need to modify multiple services to add a new thing (cloud providers, notifications, triggers, etc). If we turn this into an Any, does front50 actually need to know the internal type? I'd love a world where Echo defines a notification and all the structure that it has, but not need to have front50 change to store that structure.

front50-proto/src/main/proto/applications.proto
robzienert: OK, that's legit. I'll keep `request_id` usages isolated to long polling operations - which won't be part of this initial go-around. Thanks!

front50-proto/src/main/proto/applications.proto
robzienert: Thanks, makes sense, appreciate the clarification.

front50-proto/src/main/proto/applications.proto
plumpy: If we're saying this is Echo-owned data that's opaque to front50, I still think storing it as an Any makes more sense than a struct. Just serialize it into a byte[] and store that. Eventually Echo is going to be using protos for its internal models too, so using protos instead of maps now just inches it towards that future.

front50-proto/src/main/proto/applications.proto
plumpy: Yeah, sure, seems reasonable.

front50-proto/src/main/proto/applications.proto
brianwolfe: Ah, I think you are right @plumpy. In this case, we want the message to be opaque _except_ for the writers/readers. Because this can be completely opaque to Front50, using Any makes a lot of sense.

To clarify what I'm thinking, in Echo, we would have to have the polymorphic serialization/deserialization code, based on https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto#L60-L65:

```
// pack the slack notification spec.
SlackNotification notification = ...;
myApplication.data.put("aSlackNotifier", Any.pack(foo));

// use the deserialized form.
Any notification = myApplication.data.get("notifier")
if (any.is(SlackNotification.class)) {
  SlackNotification notification = any.unpack(Notification.class);
  notification.send(context)
}
```

Within Front50, we would never try to unpack it. Because we don't try to deserialize the encompassed envelope, we wouldn't hit any class definition errors. So in this case, Echo defines the type, uses the type, and gets all the safety involved, Front50 handles it without ever touching it.

Migrating schemas forwards still seems difficult, but easier since we at least can automatically detect backwards-incompatible schema changes.

Based on my current understanding, strongly on the `Any` side.

plumpy: For schema migration, I think you'd only do it in extreme cases. I think you could generally handle it the same as if Echo is talking to a client compiled against an old version of its protos. i.e. you added a field in Echo 2.3, but the client is using the protos from Echo 2.2. Usually you want to support that case too, although I suppose there are times when it might not be possible. Loading serialized data from an old Echo is basically the same scenario.

front50-proto/front50-proto.gradle
```diff
@@ -0,0 +1,59 @@
...
+      }
+      task.descriptorSetOptions.includeImports = true
+    }
+  }
+  generatedFilesBaseDir = "${projectDir}/build-gen/proto"
+}
+
+idea {
+  module {
+    sourceDirs += file("${projectDir}/build-gen/proto/main/java")
```
ttomsu: I saw a different example use this, which is what I used in my PR:

```
generatedSourceDirs += file("${protobuf.generatedFilesBaseDir}/main/java")
```

front50-proto/front50-proto.gradle
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+clean {
```
ttomsu: Using the `generatedSourcesDirs` above means that the build in `clean` task would work as usual - no need to for this

robzienert:
Closing - can reopen when this work picks up again.

PR #577, opened 2019-08-21
feat(s3): Front50 Storage: Support Server Side Encryption in S3 headers

justinrlee:
Supports the S3 header `x-amz-server-side-encryption-context`, with a value of either `aws:kms` or `AES256`.


justinrlee:
@ethanfrogers modified for enum ;)

PR #579, opened 2019-08-22
fix(web): Better handling when `ApplicationPermissionDAO` does not exist

ajordens:
This is only a problem as the `redis` support in `front50` does not
have support for all DAO interfaces.

Eventually i'd like to see `redis` deprecated and removed.


battlecow:
@spinnakerbot cherry-pick 1.15

PR #588, opened 2019-09-04
feat(authorization): Store resource group permissions in Front50

AbdulRahmanAlHamali:
This PR goes hand in hand with [PR-460](https://github.com/spinnaker/fiat/pull/460) in fiat to introduce the notion of resource-group level authorization. 
The PR in fiat introduces the logic of handling those groups, while this PR provides the ability to store them.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
@@ -22,12 +22,12 @@ import com.netflix.spinnaker.front50.exception.NotFoundException
 import com.netflix.spinnaker.front50.model.application.Application
 import com.netflix.spinnaker.front50.model.application.ApplicationDAO
 import com.netflix.spinnaker.front50.model.application.ApplicationPermissionDAO
+import com.netflix.spinnaker.front50.model.grouppermission.GroupPermissionDAO
 import com.netflix.spinnaker.kork.exceptions.SystemException
 import groovy.util.logging.Slf4j
 import io.swagger.annotations.ApiOperation
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
-import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
 import org.springframework.web.bind.annotation.*
 import retrofit.RetrofitError
 
@@ -41,6 +41,9 @@ public class PermissionsController {
 
   @Autowired
   ApplicationDAO applicationDAO
+
+  @Autowired
+  GroupPermissionDAO groupPermissionDAO
 
   @Autowired(required = false)
   FiatService fiatService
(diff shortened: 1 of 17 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PermissionsController.groovy
```diff
@@ -107,6 +112,36 @@ public class PermissionsController {
...
+  @RequestMapping(method = RequestMethod.POST, value = "/groups")
+  GroupPermission createGroupPermission(@RequestBody GroupPermission groupPermission) {
+    return groupPermissionDAO.create(groupPermission.getId(), groupPermission);
+  }
+
+  @ApiOperation(value = "", notes = "Update an existing group permission")
+  @RequestMapping(method = RequestMethod.PUT, value = "/groups/{groupId}")
+  GroupPermission updateGroupPermission(@PathVariable String groupId, @RequestBody GroupPermission groupPermission) {
+    groupPermissionDAO.update(groupId, groupPermission);
+    return groupPermission;
```
cfieber: we should look at what happens when we save a ServiceAccount - I think there we do a callout to fiat to re-sync roles, we probably need to do the same here

cfieber: (across create, update, and delete)

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
```diff
@@ -94,7 +94,7 @@ public class ApplicationsController {
     return results
   }
 
-  // TODO(ttomsu): Think through application creation permissions.
+  @PreAuthorize("hasPermission(#app.name, 'APPLICATION', 'CREATE')")
```
cfieber: This we need to feature flag so that we don't have the circular dependency of this depends on a new fiat, but new fiat depends on this

cfieber: This would allow feature flagging on the enforcement of `CREATE`:
```suggestion
  @PreAuthorize("#allowAllApplicationCreate or hasPermission(#app.name, 'APPLICATION', 'CREATE')")
  @ApiOperation(value = "", notes = "Create an application")
  @RequestMapping(method = RequestMethod.POST)
  Application create(@Value('${application.create.allow-all:true}') boolean allowAllApplicationCreate,
                                  @RequestBody final Application app) {
```

AbdulRahmanAlHamali:
@cfieber as mentioned in my other PR I'm considering closing this one (I'll open another one solely for adding `CREATE`), because storing the groups in front50 won't work if we want to make them configurable through halyard as with the rest of Spinnaker configuration

AbdulRahmanAlHamali:
Closing. Will store the resource groups in fiat configuration instead

PR #597, opened 2019-09-23
fix(fiat): Fix access denied for new apps with no roles

costimuraru:
Fixes https://github.com/spinnaker/spinnaker/issues/1930

Previous behavior (with LDAP enabled):
1. User creates Spinnaker app and specifies no roles.
2. Front50 does not make call to Fiat to sync permissions (this is because the roles list is empty).
3. User tries to view app.
4. Front50 makes call to Fiat to get permissions for the new app.
5. Fiat doesn't have any entry for the app (in Redis) and as such `AccessDeniedException: Access is denied` is thrown.
6. After 10 minutes Fiat will sync permissions for all apps (it has a service to sync these periodically) and the `Access denied` error will be resolved.

Behavior after this PR:
1. User creates Spinnaker app and specifies no roles.
2. Front50 makes call to Fiat to sync permissions (even though permission list is empty).
3. Fiat generates entry for the new app (in Redis).
3. User tries to view app.
4. Front50 makes call to Fiat to get permissions for the new app.
5. Fiat is able to read the entry for the new app (from Redis) and the Access denied is no longer thrown.


costimuraru:
ping @robfletcher @ttomsu @ajordens @robzienert @asher  

Relates to https://github.com/spinnaker/front50/commit/8250c54220382f8ee80b2a79abb28e74d94b09e0#diff-3d176880bdd6eb3033f247b7792c7508

costimuraru:
Good point, @robzienert!
I've added the tests now.

PR #616, opened 2019-10-25
fix(fiat-api): fix usage of removed method from Permissions

cfieber:
updates (again) to fiat 1.7.1


cfieber:
fixes the breakage that #615 addressed

PR #619, opened 2019-10-25
fix(authorization): Sync roles after an application is created

AbdulRahmanAlHamali:
Should fix https://github.com/spinnaker/spinnaker/issues/5063

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
@@ -61,7 +61,7 @@ public class ApplicationsController {
   List<ApplicationEventListener> applicationEventListeners = []
 
   @Autowired
-  FiatService fiatService;
+  Optional<FiatService> fiatService;
 
   @PreAuthorize("#restricted ? @fiatPermissionEvaluator.storeWholePermission() : true")
   @PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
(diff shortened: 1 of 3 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
```diff
@@ -98,7 +102,9 @@ public class ApplicationsController {
   @ApiOperation(value = "", notes = "Create an application")
   @RequestMapping(method = RequestMethod.POST)
   Application create(@RequestBody final Application app) {
-    return getApplication().initialize(app).withName(app.getName()).save()
+    Application createdApplication = getApplication().initialize(app).withName(app.getName()).save()
+    fiatService.sync()
```
cfieber:
```suggestion
    try {
      fiatService.ifPresent { it.sync() }
    } catch (Exception ignored) {
      log.warn("failed to trigger fiat permission sync", ignored)
    } 
```

cfieber: we've already created the application - may as well return it

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
```diff
@@ -59,6 +60,9 @@ public class ApplicationsController {
   @Autowired(required = false)
   List<ApplicationEventListener> applicationEventListeners = []
 
+  @Autowired
+  FiatService fiatService;
```
cfieber:
```suggestion
  Optional<FiatService> fiatService;
```

it is possible to run front50 without fiat

AbdulRahmanAlHamali:
Actually I think we should use [FiatStatus.isEnabled](https://github.com/spinnaker/fiat/blob/87a0d2410244f1a906a3220903aa121ab5042b24/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatStatus.java) instead of try/catch. No?

PR #621, opened 2019-10-28
feat(plugin-registry): Plugin registry with SQL implementation and protobuf as data interchange format

jonsie:
Initially gRPC was explored, but there were too many unknowns to include with this scope of work.  We explored the gRPC story around authorization, authentication, logging, metrics, exception handling - that will all need to be a separate scope of work.

This implementation uses Protobuf via REST, which should be a good pattern for us to start using as we migrate to Protobuf and eventually gRPC.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugin/PluginRepository.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugin/PluginRepository.java
@@ -1,0 +1,96 @@
+package com.netflix.spinnaker.front50.model.plugin;
+
+import java.util.List;
+
+public interface PluginRepository<T> {
+  /**
+   * Inserts or updates the specified plugin.
+   *
+   * @param data The plugin data to store
+   */
+  void upsert(UpsertData<T> data);
+
+  /**
+   * Returns the plugin by the specified name.
+   *
+   * @param criteria The criteria by which to get the plugin.
+   * @return The plugin object.
+   */
+  T get(GetCriteria criteria);
+
+  /**
+   * Get all plugins based on the criteria.
+   *
+   * @param criteria The criteria by which to filter plugins by.
+   * @return A list of matching plugins.
+   */
+  List<T> list(ListCriteria criteria);
+
+  class GetCriteria {
+    public GetCriteria(String namespace, String name) {
+      this.namespace = namespace;
+      this.name = name;
+    }
+
+    public String getNamespace() {
+      return namespace;
+    }
+
+    public void setNamespace(String namespace) {
+      this.namespace = namespace;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public void setName(String name) {
+      this.name = name;
+    }
+
+    private String namespace;
+    private String name;
+  }
+
+  class ListCriteria {
+    public ListCriteria(boolean enabled) {
+      this.enabled = enabled;
+    }
+
+    public boolean getEnabled() {
+      return enabled;
+    }
+
+    public void setEnabled(boolean enabled) {
+      this.enabled = enabled;
+    }
+
+    private boolean enabled;
+  }
+
+  class UpsertData<T> {
+    public UpsertData(T plugin, String modifiedBy) {
+      this.plugin = plugin;
+      this.modifiedBy = modifiedBy;
+    }
+
+    public T getPlugin() {
+      return plugin;
+    }
+
+    public void setPlugin(T plugin) {
+      this.plugin = plugin;
+    }
+
+    public String getModifiedBy() {
+      return modifiedBy;
+    }
+
+    public void setModifiedBy(String modifiedBy) {
+      this.modifiedBy = modifiedBy;
+    }
+
+    private T plugin;
+    private String modifiedBy;
+  }
+}
--- a/front50-protobuf/front50-protobuf.gradle
+++ b/front50-protobuf/front50-protobuf.gradle
@@ -1,0 +1,45 @@
+buildscript {
+  repositories {
+    jcenter()
+  }
+  dependencies {
+    classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.8'
+  }
+}
+
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
+
+ext {
+  protobufVersion = '3.2.+'
+}
+
+dependencies {
+  api "com.google.protobuf:protobuf-java:${protobufVersion}"
+}
+
+protobuf {
+  protoc {
+    artifact = "com.google.protobuf:protoc:$protobufVersion"
+  }
+
+  generateProtoTasks {
+    ofSourceSet('main').each { task ->
+      task.builtins {
+        java {
+          outputSubDir = 'java'
+        }
+      }
+      task.descriptorSetOptions.includeImports = true
+    }
+  }
+}
+
+sourceSets {
+  main {
+    java {
+      srcDir "${protobuf.generatedFilesBaseDir}/main/grpc"
+      srcDir "${protobuf.generatedFilesBaseDir}/main/java"
+    }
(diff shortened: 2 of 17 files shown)


jonsie:
@spinnaker-ro

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginController.java
```diff
@@ -0,0 +1,64 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.http.converter.HttpMessageConversionException;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping(value = "plugins")
+@ConditionalOnExpression("${spinnaker.plugins.enabled:false}")
+public class PluginController {
```
jonsie: Essentially, if the content type is `application/x-protobuf` this controller will produce and consume protobufs.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginController.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private final Logger log = LoggerFactory.getLogger(getClass());
+  private PluginRepository<Plugin> pluginRepository;
+
+  public PluginController(PluginRepository pluginRepository) {
+    this.pluginRepository = pluginRepository;
+  }
+
```
srekapalli: Minor.. We can remove value attr ?

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PluginControllerSpec.groovy
```diff
@@ -0,0 +1,76 @@
...
+    .addAllRequiredVersions(["clouddriver>=1", "orca>=2"])
+    .addAllDependencies(["a-necessary-thing"])
+    .setEnabled(true)
+    .setInProcessUnsafe(false)
+  Plugin plugin = pluginBuilder.build()
+  PluginRepository<Plugin> pluginRepository = Stub(PluginRepository)
+
+  @Subject
+  def controller = new PluginController(pluginRepository)
+
```
srekapalli: Probably , having some MockMVC type tests here might be helpful to exercise 404/400 type scenarios and also for the happy path that deal with protobuf msgs ?

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PluginControllerSpec.groovy
```diff
@@ -0,0 +1,180 @@
...
+    controller = new PluginController(pluginRepository)
+
+    mockMvc = MockMvcBuilders
+      .standaloneSetup(controller)
+      .setHandlerExceptionResolvers(createExceptionResolver())
+      .setMessageConverters([new ProtobufHttpMessageConverter()] as HttpMessageConverter<?>[])
+      .build()
+  }
+
```
srekapalli: 👍 💯

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugin/PluginRepository.java
```diff
@@ -0,0 +1,96 @@
+package com.netflix.spinnaker.front50.model.plugin;
+
+import java.util.List;
+
+public interface PluginRepository<T> {
```
robzienert: Include interface-level docs. What is it, what's its used for, are there any implicit behaviors that implementations should adhere to, etc? Also useful to document what `T` is or isn't, since it's totally unbounded.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugin/PluginRepository.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Get all plugins based on the criteria.
+   *
+   * @param criteria The criteria by which to filter plugins by.
+   * @return A list of matching plugins.
+   */
+  List<T> list(ListCriteria criteria);
+
+  class GetCriteria {
```
robzienert: nit: I think from an ergonomics perspective, this these criteria objects should use a builder pattern.

front50-protobuf/front50-protobuf.gradle
```diff
@@ -0,0 +1,45 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
+
+ext {
+  protobufVersion = '3.2.+'
+}
+
+dependencies {
+  api "com.google.protobuf:protobuf-java:${protobufVersion}"
```
robzienert: I _think_ the protobuf version is provided by spinnaker-dependencies? Even if it isn't, I think it should be so we don't have individual repos defining this.

front50-protobuf/src/main/proto/com/netflix/spinnaker/front50/proto/plugins.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+
+option java_package = "com.netflix.spinnaker.front50.proto";
+option java_multiple_files = true;
+
+message Plugin {
```
robzienert: Need more docs on both the message-level and the individual property level. For example, `version` - this should be known that it adheres to semver.org.

front50-protobuf/src/test/resources/test.txt
```diff
@@ -0,0 +1,5 @@
+cat src/test/resources/event.txt | protoc --encode "Plugin" src/main/proto/com/netflix/spinnaker/front50/proto/plugins.proto | curl -H "X-SPINNAKER-USER: person@example.com" -H "Content-Type: application/x-protobuf" -X POST http://localhost:7101/plugins --data-binary @-
+
+curl -H "X-SPINNAKER-USER: person@example.com" -H "Content-Type: application/x-protobuf" -X GET http://localhost:7101/plugins/somePlugin/netflix
+
+curl -H "X-SPINNAKER-USER: person@example.com" -H "Content-Type: application/x-protobuf" -X GET http://localhost:7101/plugins
```
robzienert: Could use a better filename, here. Testing what?

front50-sql/src/main/resources/db/changelog/20191025-initial-plugins-schema.yml
```diff
@@ -0,0 +1,94 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-plugins-table
+      author: csmalley
```
robzienert: Use your github name.

front50-sql/src/test/kotlin/com/netflix/spinnaker/front50/model/plugin/SqlPluginRepositoryTests.kt
```diff
@@ -0,0 +1,132 @@
...
+
+      test("Able to get a plugin") {
+        val plugin = pluginBuilder.build()
+        sqlPluginRepository.upsert(UpsertData(plugin, "someone"))
+
+        val result = sqlPluginRepository.get(GetCriteria(plugin.namespace, plugin.name))
+
+        expect {
+          assert(result == plugin)
+        }
```
robzienert: `expect(result).isEqualTo(plugin)`

jonsie:
Well, I wrote all this and then realized there's `StorageServiceSupport` in front50. 
🔥 So, I'm burning this down.

That model supports multiple backends out of the box, which is nice.  However, that is strictly implemented with POJO serialization/deserialization for JSON.  Adding Protobuf support is going to be a chunk of work.  So, I'm going to close this PR and do things the normal way, and we're gonna use JSON.

PR #624, opened 2019-10-31
feat(Postgres DB Storage): Adding Postgres SQL storage dependencies

srekapalli:
- Adding dependencies to support Postgres DB as an option for Storage.
- Storage service tests now go against both mysql and postgres test containers.
- Had to tweak upsert API so that both MySQL and Postgres will work.


front50-sql/src/main/resources/db/changelog/20190530-add-is-deleted-indexes.yml
```diff
@@ -103,3 +107,111 @@ databaseChangeLog:
       - dropIndex:
           indexName: is_deleted_idx
           tableName: snapshots
+  - changeSet:
+      id: create-indexes-postgresql
+      author: ajordens
```
cfieber: should update author

front50-sql/src/test/kotlin/com/netflix/spinnaker/front50/model/SqlStorageServiceTests.kt
```diff
@@ -23,235 +23,253 @@ import com.netflix.spinnaker.front50.model.application.Application
...
+
+          expectThat(
+            jooq
+              .select(DSL.field("id", String::class.java))
+              .from("pipelines")
+              .where(
+                DSL.field("name", String::class.java).eq("pipeline001_updated")
+              )
+              .fetchOne(DSL.field("id", String::class.java))
+          ).trim().isEqualTo("id-pipeline001")
```
cfieber: why is the trim needed here?

cfieber: (was specifically added in this changeset)

srekapalli: Was receiving string with length equal to size of the column when I run against a Test container. This is not a problem with the real DB.

cfieber: can you confirm that is also the case with real postgres (vs cockroach)?  testcontainer should be running postgres..

srekapalli: Good callout.  Postgres TestContainer  and actual postgres are behaving the same way.  Liquibase is converting char datatype to bpchar (blankpadded char) and comparisons are not working when using text of arbitrary length.  Postgres preferred datatype is VARCHAR instead of CHAR ,I will adjust liquibase to use the new type.

cfieber:
is the plan still to add the liquibase database changeset to migrate the columns to varchar?

srekapalli:
> is the plan still to add the liquibase database changeset to migrate the columns to varchar?

For now, will leave MySQL changeset as is.  Will do that in a different PR ?  Atleast with this change, current OSS versions running on MySQL should have no impact.

cfieber:
my only concern is the `trim()` in the test masking a known problem (as it stands this is front50-sql-cockroach not front50-sql-postgres). If the follow up PR will be pretty immediate then I'm ok with this going in, otherwise lets get some documentation in the module describing the current state of stability/support (README.md or something).

srekapalli:
O.. sorry. Forgot to mention. **I removed trim** .. Now the changes are basically, 

-- New changesets targeting postgres.
-- Tests run with no changes against postgres & mysql
-- Postgres uses varchar and for mysql will put in another PR adjusting datatype CHAR to VARCHAR.

front50-sql/src/main/resources/db/changelog/20190415-initial-application-permissions-schema.yml
```diff
@@ -2,6 +2,10 @@ databaseChangeLog:
   - changeSet:
       id: create-application-permissions-table
       author: ajordens
+      preConditions:
+        - onFail: CONTINUE
+        - dbms:
+            type: mysql
```
robzienert: Modifications to an existing changelog will alter its checksum, subsequently causing migrations to fail and thus breaking future deployments of the service - how will existing users of SQL in Front50 upgrade to this version once merged?

robzienert: Are these `preConditions` fields meant to address this?

Do you have any thoughts / guidance for keeping schema changes in-sync across dialects when alterations are needed in the future?

front50-sql/src/main/resources/db/changelog/20190415-initial-pipelines-schema.yml
```diff
@@ -75,7 +83,7 @@ databaseChangeLog:
                     nullable: false
               - column:
                   name: body_sig
-                  type: char(32)
+                  type: varchar(32)
```
robzienert: `char` is standard SQL. Is there an actual issue with this one being `char(32)`? The above change around `char(255) -> varchar(255)` makes sense, but a signature column is going to be fixed length.

front50-sql/src/main/resources/db/changelog/20190415-initial-application-permissions-schema.yml
srekapalli: > Modifications to an existing changelog will alter its checksum, subsequently causing migrations to fail and thus breaking future deployments of the service - how will existing users of SQL in Front50 upgrade to this version once merged?

Verified that preconditions aren't part of the checksum , so we should be OK.

http://www.liquibase.org/2009/03/what-effects-changeset-checksums.html

srekapalli:
I'm closing this PR. Will add a incremental changeset starting with what we have now and apply postgres changes on top of those conditionally.

PR #625, opened 2019-10-31
feat(plugin-registry): Front50 plugins storage service

jonsie:
Plain ol' JSON common storage service front50 implementation.  Protobuf needs a bit of extra work to get it working with this implementation.  

Previous PR:  https://github.com/spinnaker/front50/pull/621

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/DefaultPluginRepository.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/DefaultPluginRepository.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.front50.model.plugin;
+
+import com.google.common.hash.Hashing;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.front50.exception.NotFoundException;
+import com.netflix.spinnaker.front50.model.ObjectKeyLoader;
+import com.netflix.spinnaker.front50.model.ObjectType;
+import com.netflix.spinnaker.front50.model.StorageService;
+import com.netflix.spinnaker.front50.model.StorageServiceSupport;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import rx.Scheduler;
+
+public class DefaultPluginRepository extends StorageServiceSupport<Plugin>
+    implements PluginRepository {
+
+  public DefaultPluginRepository(
+      StorageService service,
+      Scheduler scheduler,
+      ObjectKeyLoader objectKeyLoader,
+      long refreshIntervalMs,
+      boolean shouldWarmCache,
+      Registry registry) {
+    super(
+        ObjectType.PLUGIN,
+        service,
+        scheduler,
+        objectKeyLoader,
+        refreshIntervalMs,
+        shouldWarmCache,
+        registry);
+  }
+
+  @Override
+  public Collection<Plugin> list(boolean enabled) {
+    return all().stream()
+        .filter(plugin -> plugin.isEnabled() == enabled)
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public Plugin get(String namespace, String name) throws NotFoundException {
+    String id = Hashing.sha256().hashString(namespace + name, StandardCharsets.UTF_8).toString();
+    return all().stream()
+        .filter(plugin -> plugin.getId().equalsIgnoreCase(id))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new NotFoundException(
+                    String.format(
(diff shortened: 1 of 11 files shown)


front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -19,17 +19,18 @@ package com.netflix.spinnaker.front50.model
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.front50.exception.NotFoundException
-import com.netflix.spinnaker.front50.model.ObjectType.APPLICATION
-import com.netflix.spinnaker.front50.model.ObjectType.APPLICATION_PERMISSION
-import com.netflix.spinnaker.front50.model.ObjectType.DELIVERY
-import com.netflix.spinnaker.front50.model.ObjectType.ENTITY_TAGS
-import com.netflix.spinnaker.front50.model.ObjectType.NOTIFICATION
+import com.netflix.spinnaker.front50.model.ObjectType.PROJECT
```
cfieber: interesting re-sort of the imports here - guessing that was spotless being.. helpful?

front50-sql/src/main/resources/db/changelog/20191030-initial-plugins-schema.yml
```diff
@@ -0,0 +1,47 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: char(255)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: body
+                  type: longtext
```
cfieber: my sql-neckbeard isn't fabulous but I think it's better to stick a longtext column at the end of the columns list? someone else probably knows better

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/Plugin.java
```diff
@@ -0,0 +1,145 @@
...
+
+package com.netflix.spinnaker.front50.model.plugin;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.common.hash.Hashing;
+import com.netflix.spinnaker.front50.model.Timestamped;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+public class Plugin implements Timestamped {
```
cfieber: what is the story for the location of the plugin binary?

I like that it isn't directly in here, but should this include a pointer URI or something to the contents?

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
jonsie: Yes... spotless.  I will sort this lexically though, that'll be better.

front50-sql/src/main/resources/db/changelog/20191030-initial-plugins-schema.yml
jonsie: I suppose it can't hurt.  This is a copy paste of other liquibase migration files for a generic object blobs (for example, the exact same as `snapshots`, `notifications`, etc).  But yea, I'll do this.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/Plugin.java
jonsie: I think that is an open question.  I'm going to hold off on adding that for the time being.

jonsie: Expanding on that - I believe the current thought was to specify repositories via kork.

cfieber: 👍 seems legit to start with

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/DefaultPluginRepository.java
```diff
@@ -0,0 +1,93 @@
...
+  @Override
+  public Plugin upsert(Plugin plugin) {
+    return create(plugin.getId(), plugin);
+  }
+
+  @Override
+  public Plugin create(String id, Plugin plugin) {
+    plugin.setLastModified(System.currentTimeMillis());
+
+    Assert.notNull(plugin.getName(), "Plugin name must NOT be null!");
```
cfieber: I'd prefer Objects.requireNonNull (same functionality, but jdk native)

jonsie:
> Do any of the other storage implementations besides SQL need to be updated to make this work? (I think no..)
> 
> I suspect we want a ConditionalOnProperty as to whether to expose this controller (I know we did that most recently on the DeliveryController

From what I can tell the non-SQL implementations should just work.  I'll add a `ConditionalOnProperty` to the `PluginController`.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugin/DefaultPluginRepository.java
jonsie: Can do, I think I'll catch the NPEs and throw an `IllegalArgumentException` though.

front50-sql/src/main/resources/db/changelog/20191030-initial-plugins-schema.yml
```diff
@@ -0,0 +1,47 @@
...
+  - changeSet:
+      id: create-plugins-table
+      author: jonsie
+      changes:
+        - createTable:
+            tableName: plugins
+            columns:
+              - column:
+                  name: id
+                  type: char(255)
```
robzienert: The `id` property on plugins is sha256. This should be `char(64)`.

jonsie: Good catch.

jonsie:
Is this necessary?

It may be too soon (or not necessary) to add this functionality to Front50, or it may be that this manifest can be greatly simplified.

robzienert:
Closing for now.

PR #635, opened 2019-11-13
feat(PostgreSQL DB Storage): Added support for Postgres. Changes incl…

srekapalli:
- Added changesets for postgres 
- Master changeset will include changesets required for postgres which get conditionally executed based on the `dbms` type.
- Modified an existing changeset ( for adding indexes ) to conditionally execute for MySQL only.
- Adjusted JOOQ api to work with both MySQL and PostgreSQL.

/cc @cfieber @robzienert @ajordens


front50-sql/src/main/resources/db/changelog/20190530-add-is-deleted-indexes-postgres.yml
```diff
@@ -0,0 +1,109 @@
+databaseChangeLog:
+  - changeSet:
+      preConditions:
+        onFail: CONTINUE
+        dbms:
+          type: postgresql
+      id: create-indexes-postgresql
```
srekapalli: This changeset is similar to the MySQL one except that PgSQL indexes are globally scoped and we can't have same index name for every table and this changeset is applying that rule.

front50-sql/src/main/resources/db/changelog/20190530-add-is-deleted-indexes.yml
```diff
@@ -1,5 +1,9 @@
 databaseChangeLog:
```
srekapalli: This existing changeset will only be applied to MySQL DB.  Adding Preconditions will not change the checksum.  Verified this.

http://www.liquibase.org/2009/03/what-effects-changeset-checksums.html

front50-sql/src/main/resources/db/changelog/20191112-modify-char-columns-postgres.yml
```diff
@@ -0,0 +1,89 @@
+databaseChangeLog:
```
srekapalli: This is a new changeset only applied for Postgres.  CHAR types are converted to bpchar (blank pad characters) in Postgres and they are not recommended datatypes + JOOQ doesn't play well with padded column values. So this change set modifies the column data type to VARCHAR.

robzienert: I did a quick look around but came up short - could you link where it's documented that `char` / `bpchar` is not recommended? I'd like to get more learned on this.

front50-sql/src/main/resources/db/changelog/20190530-add-is-deleted-indexes-postgres.yml
robzienert: What happens with the schemas defined previously? Does pg still create the indexes but they're wrong, or what? I would expect that if the indexes conflict, it would generate an error and fail to apply the changeset, so we'd never land here?

srekapalli: PG will fail to execute the changeset.

srekapalli: Not sure what you meant by `so we'd never land here` , but our original changeset that we had will fail for postgres and that's the main reason I had to add preconditions in there.

front50-sql/src/main/resources/db/changelog/20191112-modify-char-columns-postgres.yml
srekapalli: https://stackoverflow.com/questions/51528982/what-comparison-problems-can-using-the-varchar-type-instead-of-the-char-type-avo

front50-sql/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+#### PostgreSQL:
+```yaml
+sql:
+  enabled: true
+  baseUrl: jdbc:postgresql://localhost:5432
+  connectionPools:
+    default:
+      jdbcUrl: ${sql.baseUrl}/front50?useSSL=false&serverTimezone=UTC
```
cfieber:
```suggestion
      jdbcUrl: ${sql.baseUrl}
```

front50-sql/README.md
```diff
@@ -0,0 +1,35 @@
...
+    jdbcUrl: ${sql.baseUrl}/front50?useSSL=false&serverTimezone=UTC
+    user: 
+    password:
+```
+
+#### PostgreSQL:
+```yaml
+sql:
+  enabled: true
+  baseUrl: jdbc:postgresql://localhost:5432
```
cfieber:
```suggestion
  baseUrl: jdbc:postgresql://localhost:5432/front50
```

front50-sql/README.md
```diff
@@ -0,0 +1,35 @@
...
+  connectionPools:
+    default:
+      jdbcUrl: ${sql.baseUrl}/front50?useSSL=false&serverTimezone=UTC
+      dialect: POSTGRES
+      user: 
+      password:
+  migration:
+    jdbcUrl: ${sql.baseUrl}/front50?useSSL=false&serverTimezone=UTC
```
cfieber:
```suggestion
    jdbcUrl: ${sql.baseUrl}
```

front50-sql/README.md
cfieber: those query args are mysql specific (postgres is `ssl` vs `useSSL` but also defaults to false)

PR #641, opened 2019-12-02
feat(sql): Enables composite storage service to be backed by SQL stor…

srekapalli:
Enables composite storage service to be backed by 2 SQL storage services. 
In order to wire up with storage services of same type, configuration now accepts bean names.
To target different pools for any CRUD operations , all storage service methods are wrapped with **`withPool`** 
Added test sample config to drive the test with 2 SQL storage service impls.

### DIFF
--- a/front50-sql/front50-sql.gradle
+++ b/front50-sql/front50-sql.gradle
@@ -26,6 +26,8 @@ dependencies {
   testImplementation "org.testcontainers:postgresql"
   testImplementation "org.postgresql:postgresql"
 
+  testImplementation "com.h2database:h2"
+
 }
 
 test {
--- a/front50-sql/front50-sql.gradle
+++ b/front50-sql/front50-sql.gradle
@@ -26,6 +26,7 @@ dependencies {
   testImplementation "org.testcontainers:postgresql"
   testImplementation "org.postgresql:postgresql"
 
+  // Only used for Initializing Datasource. For actual CRUD, test containers preferred.
   testImplementation "com.h2database:h2"
 
 }
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
@@ -23,6 +23,8 @@ import com.netflix.spinnaker.front50.model.SqlStorageService
 import com.netflix.spinnaker.kork.sql.config.DefaultSqlConfiguration
 import com.netflix.spinnaker.kork.sql.config.SqlProperties
 import org.jooq.DSLContext
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.annotation.Bean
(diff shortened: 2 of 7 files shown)


front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -39,14 +41,35 @@ class SqlConfiguration : CommonStorageServiceDAOConfig() {
...
+    )
+
+  @Bean
+  @ConditionalOnProperty("sql.secondary.enabled", matchIfMissing = false)
+  fun secondarySqlStorageService(
+    objectMapper: ObjectMapper,
+    registry: Registry,
+    @Qualifier("secondaryJooq") jooq: DSLContext,
+    sqlProperties: SqlProperties,
+    @Value("\${sql.secondary.pool-name}") poolName: String
```
jonsie: I think it would be cleaner to define both `sql.primary.pool-name` and `sql.secondary.pool-name` in a class annotated with `@ConfigurationProperties`.   Informally, I think we're trying to steer clear of introducing new `@Value` annotations.

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -55,7 +56,8 @@ class SqlStorageService(
   private val jooq: DSLContext,
   private val clock: Clock,
   private val sqlRetryProperties: SqlRetryProperties,
-  private val chunkSize: Int
+  private val chunkSize: Int,
+  private val poolName: String = "default"
```
jonsie: I think if we define `poolName` via a configuration properties class, we can assign the default value there and pass it into the constructor and then remove the default value here.

front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
srekapalli: Instead of introducing more props, derived the pool name using the existing connection pool config attributes.  This way we don't have to duplicate the name into yet another prop.

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
srekapalli: Took the same approach here, derive the default pool name from the existing connection pool settings.

front50-sql/front50-sql.gradle
```diff
@@ -26,6 +26,8 @@ dependencies {
   testImplementation "org.testcontainers:postgresql"
   testImplementation "org.postgresql:postgresql"
 
+  testImplementation "com.h2database:h2"
```
cfieber: I think this is okay for now but we've trended away from h2 towards testcontainers for most of the SQL testing. I think the use case you have this here for is okay and makes sense (just initing storage services) but wouldn't want to see anything beyond that using h2.

jonsie: In that case, I'd like to see a comment to that nature here.

srekapalli: Done!!

PR #644, opened 2019-12-05
feat(core): Add endpoint to get current pipeline definition

ezimanyi:
There's currently no endpoint in front50 to get the current pipeline definition; consumers need to call the history endpoint with a limit of 1 and take the single returned element.

Let's make the API simpler for consumers who just want the current definition by giving them an endpoint to return it.

Forcing consumers to call the /history endpoint has led to performance issues, as that endpoint is often much slower than a call to get the current definition. Some backing stores have optimized to short-circuit the limit=1 case (s3 in #316), but others (including GCS) are still very slow even in the limit=1 case.

We could (and potentially should) make the same optimization to other backing stores. But I think the fundamental issue here is that we have rare, slow history-type requests and frequent (ideally) fast
current-state requests going to the same endpoint. This PR creates that distinction, and consumers can elect to use the new endpoint going forward.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy
@@ -29,6 +29,7 @@ import com.netflix.spinnaker.kork.web.exceptions.ValidationException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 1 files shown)


ezimanyi:
This is a first step in solving the general performance problem noted in [this comment](https://github.com/spinnaker/spinnaker/issues/5085#issuecomment-562125988).  Once this is in place, calls to front50 that don't need history can use this new endpoint.

dreynaud:
Thanks @ezimanyi, I ❤️ this

PR #651, opened 2019-12-10
feat(plugins): Adds plugin artifact repository

robzienert:
This is an optional feature for plugins. This makes front50 the source of
truth of what plugins are available for download by any particular service.
Which plugin release is actually downloaded for a particular service is
determined by the service itself.

⚠️ Going to hold off on merge until I write up the `Front50UpdateRepository` in `kork-plugins`, just getting all of the pieces in place.

Before this is merged, there's question about how much context we put into the repository on initial roll out. Right now I have things written such that plugin releases define what services they support, but nothing much beyond that. Should plugin releases also define service version constraints? I think yes, but I'm unsure how much we need to bite off out of the gate.

This PR just provides services of knowledge on what plugins to download, and where to download them from. It does not provide functionality around plugin downloads or configuration of downloaded plugins.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/DefaultPluginArtifactRepository.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/DefaultPluginArtifactRepository.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.front50.model.pluginartifact;
+
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.front50.model.ObjectKeyLoader;
+import com.netflix.spinnaker.front50.model.ObjectType;
+import com.netflix.spinnaker.front50.model.StorageService;
+import com.netflix.spinnaker.front50.model.StorageServiceSupport;
+import com.netflix.spinnaker.kork.exceptions.IntegrationException;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import rx.Scheduler;
+
+public class DefaultPluginArtifactRepository extends StorageServiceSupport<PluginArtifact>
+    implements PluginArtifactRepository {
+  public DefaultPluginArtifactRepository(
+      StorageService service,
+      Scheduler scheduler,
+      ObjectKeyLoader objectKeyLoader,
+      long refreshIntervalMs,
+      boolean shouldWarmCache,
+      Registry registry) {
+    super(
+        ObjectType.PLUGIN_ARTIFACT,
+        service,
+        scheduler,
+        objectKeyLoader,
+        refreshIntervalMs,
+        shouldWarmCache,
+        registry);
+  }
+
+  @Nonnull
+  @Override
+  public Collection<PluginArtifact> getByService(@Nonnull String service) {
+    return all().stream()
+        .filter(
+            artifact -> artifact.getReleases().stream().anyMatch(r -> r.supportsService(service)))
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public PluginArtifact create(String id, PluginArtifact item) {
+    Objects.requireNonNull(item.getId());
+    if (!item.getId().equals(id)) {
+      // Won't happen unless Orca passes a mismatched request path / request body.
+      throw new IntegrationException("The provided id and plugin artifact id do not match");
+    }
+
+    if (item.getCreateTs() == null) {
+      item.setCreateTs(System.currentTimeMillis());
+    } else {
+      item.setLastModified(System.currentTimeMillis());
+    }
+
+    update(id, item);
+    return findById(id);
+  }
+}
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifact.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifact.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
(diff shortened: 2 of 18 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifact.java
```diff
@@ -0,0 +1,129 @@
...
+     */
+    private String version;
+
+    /** The date of the plugin artifact release. */
+    private String date;
+
+    /**
+     * Spinnaker service support.
+     *
+     * <p>This defines, from the plugin's perspective, what services it supports. This wil be used
```
claymccoy: wil > will

claymccoy:
What exactly does it mean for a plugin to be active? This is a Spinnaker rather than a PF4J-update attribute right? Can apps only use active releases of a plugin? If so, it does seem restrictive to only allow one to be active. I read the comment about introducing the concept of plugin stability, and that makes sense.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifact.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import lombok.Data;
+
+/**
+ * A Spinnaker plugin's release artifact metadata.
+ *
+ * <p>This model is used internally by Spinnaker to track what plugins are available for services to
+ * install, as well as the specific releases that should be installed.
+ */
```
srekapalli: New entity requires Liquibase migration.?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/DefaultPluginArtifactRepository.java
```diff
@@ -0,0 +1,79 @@
...
+  public Collection<PluginArtifact> getByService(@Nonnull String service) {
+    return all().stream()
+        .filter(artifact -> hasActiveReleaseMatchingService(artifact, service))
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public PluginArtifact create(String id, PluginArtifact item) {
+    Objects.requireNonNull(item.getId());
+    if (!item.getId().equals(id)) {
```
srekapalli: very minor. Is the comment about Orca relevant here ..?

srekapalli:
LGTM. I'm assuming we will also add a controller/endpoint for CRUD .?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifact.java
robzienert: Yep - added that now.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/DefaultPluginArtifactRepository.java
robzienert: Yes, it is. The Orca bits haven't been done yet, but they will need to be.

robzienert:
@claymccoy I've updated the PR, thanks for your confirmation that `active` wasn't good enough. Changes:

- A `Release`'s `requires` field now takes a `{service}{comparator}{semVer}` format, e.g. `clouddriver>=1.0.0`
- A `Release` now has a `state` enum (e.g. `CANDIDATE`, `STABLE`). Is there a better word for this than `state`?

With these two pieces of metadata, the service will have information to make some reasonable choices of what plugin version to install. I suspect we'll need more in the future for blacklisting / deprecating releases - but that can be added on later.

robzienert:
@srekapalli Controller added, plus some other things.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pluginartifact/PluginArtifactService.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private void validate(PluginArtifact pluginArtifact) {
+    Errors errors = new GenericValidationErrors(pluginArtifact);
+    validators.forEach(v -> v.validate(pluginArtifact, errors));
+    if (errors.hasErrors()) {
+      throw new ValidationException(errors);
+    }
+  }
+
```
srekapalli: Do we need  exception handler serializing this exception or we have a generic exception handler somewhere in Kork  ?

robzienert: We have a generic handler: https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginArtifactController.java
```diff
@@ -0,0 +1,77 @@
...
+        .map(pluginArtifactService::findAllByService)
+        .orElseGet(pluginArtifactService::findAll);
+  }
+
+  @RequestMapping(value = "", method = RequestMethod.POST)
+  PluginArtifact create(@RequestBody PluginArtifact pluginArtifact) {
+    return pluginArtifactService.upsert(pluginArtifact);
+  }
+
+  @RequestMapping(value = "/{id}", method = RequestMethod.POST)
```
jonsie: The `id` is not passed to `upsert`, and even if it were passed to `upsert` -- if the `id` doesn't exist the upsert creates the plugin anyways.  

Do we want to provide a pattern for passing a known plugin ID and plugin artifact?  I'm not really sure I understand the point of this endpoint.

robzienert: Ha. You're right. I'm going to keep the endpoint (since it follows convention), but just assert they're the same.

robzienert: Jk, that's not what I'm doing.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginArtifactController.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+/** TODO(rz): What's the permissions model for something like artifacts? */
+@RestController
+@RequestMapping("/pluginArtifacts")
+public class PluginArtifactController {
```
jonsie: Should we add a `GET` plugin artifact by ID?  I'm thinking of the context wherein we want to assert a plugin exists before calling delete.

PR #660, opened 2019-12-13
feat(chaosmonkey): Enforce chaos monkey permissions when enabled

robzienert:
Enforces that Chaos Monkey has the permissions it needs to terminate instances within an application if enabled. This introduces a new config: `spinnaker.chaos-monkey.user-role`, which should be the principal user that Chaos Monkey works as (e.g. `chaos-monkey@example.com`. Updates to either application settings or application permissions will trigger this event listener.

While working on this, I also moved business logic out of the controller layer and into a service.

The PR is broken into two commits, the first is just moving logic for permissions into `ApplicationPermissionsService`, the second is for adding the chaos monkey functionality.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.front50.listeners;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.front50.config.ChaosMonkeyEventListenerConfigurationProperties;
+import com.netflix.spinnaker.front50.events.ApplicationEventListener;
+import com.netflix.spinnaker.front50.events.ApplicationPermissionEventListener;
+import com.netflix.spinnaker.front50.model.application.Application;
+import com.netflix.spinnaker.front50.model.application.ApplicationDAO;
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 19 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -0,0 +1,178 @@
...
+    if (!(rawPermissions instanceof Map)) {
+      // Exit early, no permissions config found on the application
+      log.warn("No permissions config found on application '{}'", updatedApplication.getName());
+      return updatedApplication;
+    }
+
+    @SuppressWarnings("unchecked")
+    Map<String, List<String>> permissionsMap = (Map<String, List<String>>) rawPermissions;
+
+    if (isChaosMonkeyEnabled) {
```
cfieber: should exit early here if the permissionsMap is empty, think we serialize empty permissions as `{}`

front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -0,0 +1,178 @@
...
+      // Chaos Monkey is enabled, apply permissions
+      List<String> readPermissions = permissionsMap.get("READ");
```
cfieber:
```suggestion
      List<String> readPermissions = permissionsMap.computeIfAbsent("READ", k -> new ArrayList<>());
```
in case there is no specific entry for that role

front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -0,0 +1,178 @@
...
+
+    return updatedPermission;
+  }
+
+  private void applyNewPermissions(Application.Permission updatedPermission, boolean addRole) {
+    Permissions permissions = updatedPermission.getPermissions();
+
+    Map<Authorization, List<String>> data = new HashMap<>();
+    for (Authorization auth : Authorization.values()) {
+      List<String> roles = new ArrayList<>(permissions.get(auth));
```
jonsie: I notice we do this type of unpacking and rebuilding of Permissions in many places.  Maybe we should add a public method on Permissions like this:

```
  public Map<Authorization, List<String>> unpack() {
    return Arrays.stream(Authorization.values()).collect(toMap(identity(), this::get));
  }
```

So we could just call `permissions.unpack()`.

jonsie: Yea or nay? https://github.com/spinnaker/fiat/pull/532

front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
robzienert: Is an empty permissions map indicative of something? Does that basically mean that everyone has access?

jonsie: Yes, it means the resource is "unrestricted".

PR #672, opened 2020-01-09
fix(permissions): Update and persist chaos monkey permissions on application PRE_UPDATE event

jonsie:
The previous implementation was flawed for a few reasons.

1) ApplicationDAO does not persist permissions, so the implementation of `ApplicationEventListener` did nothing.
2) When a user updates Chaos Monkey permissions from Deck, permissions are not passed to the Orca task `UpsertApplicationTask`, which means that the chaos monkey `ApplicationPermissionEventListener` implementation would not be triggered.
3) When the chaos monkey `ApplicationPermissionEventListener` was triggered (via an Application permissions update event, like saving perms through Deck), a check to see if the permissions were unrestricted did not exist.  This would lock people out of READ and WRITE permissions since only the chaos monkey permission would be applied.

This fix addresses these issues.  I've removed the implementation `ApplicationPermissionEventListener`, and instead we use the `ApplicationPermissionsService` in the `ApplicationEventListener` implementation to ensure permissions are persisted.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
@@ -18,17 +18,13 @@ package com.netflix.spinnaker.front50.listeners;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.fiat.model.Authorization;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.front50.ApplicationPermissionsService;
 import com.netflix.spinnaker.front50.config.ChaosMonkeyEventListenerConfigurationProperties;
 import com.netflix.spinnaker.front50.events.ApplicationEventListener;
-import com.netflix.spinnaker.front50.events.ApplicationPermissionEventListener;
 import com.netflix.spinnaker.front50.model.application.Application;
-import com.netflix.spinnaker.front50.model.application.ApplicationDAO;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -64,47 +59,28 @@ public boolean supports(ApplicationEventListener.Type type) {
...
-      return updatedApplication;
-    }
+    Application.Permission permission =
+        applicationPermissionsService.getApplicationPermission(updatedApplication.getName());
 
-    Object rawPermissions = updatedApplication.details().get("permissions");
-    if (!(rawPermissions instanceof Map)) {
-      // Exit early, no permissions config found on the application
-      log.warn("No permissions config found on application '{}'", updatedApplication.getName());
+    if (!permission.getPermissions().isRestricted()) {
```
alanmquach: Will this cover the case where my application was restricted and had chaos enabled and I now want to remove the restriction? Basically going from `me` to `` (from the UI's perspective).

If so, perhaps a test permutation for that would be useful?

jonsie: Good catch.  I added a check for this, including some tests.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -176,6 +122,23 @@ private boolean isChaosMonkeyEnabled(Application application) {
...
+        .contains(properties.getUserRole());
+  }
+
+  private boolean onlyChaosMonkeyPermissions(
+      Application.Permission updatedPermission, Authorization authorizationType) {
+    return updatedPermission.getPermissions().get(authorizationType).stream()
+            .filter(role -> role.equals(properties.getUserRole()))
+            .distinct()
+            .count()
+        == 1;
```
jonsie: I think this is spotless trying to be helpful and I don't like it.

PR #677, opened 2020-01-10
fix(permissions): Only remove chaos monkey if it is the only permission, plus small code cleanup

jonsie:
In my glee to make sure we never ended up in a situation that locked people out, I got a bit carried away with removing permissions.  I really hope this is the final solution here.  Also includes a small code cleanup.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyApplicationPermissionEventListener.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyApplicationPermissionEventListener.java
@@ -69,11 +69,7 @@ public class ChaosMonkeyApplicationPermissionEventListener extends ChaosMonkeyEv
 
     Application application = applicationDAO.findByName(updatedPermission.getName());
 
-    if (isChaosMonkeyEnabled(application)) {
-      applyNewPermissions(updatedPermission, true);
-    } else {
-      applyNewPermissions(updatedPermission, false);
-    }
+    applyNewPermissions(updatedPermission, isChaosMonkeyEnabled(application));
 
     return updatedPermission;
   }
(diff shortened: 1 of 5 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyEventListener.java
```diff
@@ -80,6 +83,13 @@ private boolean shouldAddChaosMonkeyPermission(
         && !updatedPermission.getPermissions().get(authorizationType).isEmpty();
   }
 
+  /** We only want to remove chaos monkey permissions if it is the only permission. */
+  private boolean shouldRemove(
```
alanmquach: ❤️ these methods. makes it super clear how whether to add/remove is determined

front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners/ChaosMonkeyApplicationPermissionEventListener.java
```diff
@@ -69,11 +69,7 @@ public boolean supports(ApplicationPermissionEventListener.Type type) {
 
     Application application = applicationDAO.findByName(updatedPermission.getName());
 
-    if (isChaosMonkeyEnabled(application)) {
-      applyNewPermissions(updatedPermission, true);
-    } else {
-      applyNewPermissions(updatedPermission, false);
-    }
+    applyNewPermissions(updatedPermission, isChaosMonkeyEnabled(application));
```
alanmquach: nitpick: would it make sense to have `applyNewPermissions` be responsible for calling `isChaosMonkeyEnabled`?

jonsie: Yeah, if I get back to this I'll make that happen.

PR #690, opened 2020-01-20
fix(plugins): Use kork-plugins VersionRequirementsParser and CanonicalPluginId in plugin info validators

jonsie:
Depends on https://github.com/spinnaker/kork/pull/480

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/HasCanonicalPluginIdValidator.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/HasCanonicalPluginIdValidator.java
@@ -16,18 +16,16 @@
 package com.netflix.spinnaker.front50.validator;
 
 import com.netflix.spinnaker.front50.model.plugininfo.PluginInfo;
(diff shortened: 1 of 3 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/validator/HasCanonicalPluginIdValidator.java
```diff
@@ -16,18 +16,16 @@
...
 
 @Component
 public class HasCanonicalPluginIdValidator implements PluginInfoValidator {
 
-  private final Pattern pattern = Pattern.compile("[\\w]+\\.[\\w]+");
-
   @Override
   public void validate(PluginInfo pluginInfo, Errors validationErrors) {
-    if (!pattern.matcher(pluginInfo.getId()).matches()) {
+    if (!CanonicalPluginId.Companion.isValid(pluginInfo.getId())) {
```
jonsie: This was the impetus for the change, but I figured we may as well use kork-plugins for both validators.

PR #714, opened 2020-02-18
feat(plugins): Plugin binary store service

robzienert:
Refactors `front50-s3` config so that the StorageService can decoupled from S3 configuration. Added a new storage service just for plugin binaries that can use different (already available) blob stores. Only wrote an S3 implementation for now since that's all Netflix cares about.

Also wrote a little local filesystem cache that delegates to actual storage services.

Additional steps:

1. Get the gradle plugin to work with these endpoints
2. Get the plugin framework to download from front50

For Front50 plugin bootstrapping itself, its plugins will need to be baked into the image, or sourced from an already existing Front50 deployment.

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/plugins/S3PluginBinaryStorageService.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/plugins/S3PluginBinaryStorageService.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


jonsie:
I like the S3 config refactors, much cleaner.

front50-s3/src/main/java/com/netflix/spinnaker/front50/plugins/S3PluginBinaryStorageService.java
```diff
@@ -0,0 +1,104 @@
...
+    this.properties = properties;
+  }
+
+  @Override
+  public void store(@Nonnull String key, @Nonnull byte[] item) {
+    ObjectMetadata metadata = new ObjectMetadata();
+    metadata.setContentLength(item.length);
+    metadata.setContentMD5(Hashing.md5().hashBytes(item).toString());
+
+    amazonS3.putObject(
```
jonsie: What happens here if the binary exists?

robzienert: Oh, it throws an exception. I'm surprised you didn't see this. 🤐

PR #715, opened 2020-02-19
fix(s3): Refactor S3 config to support separate plugin binaries bucket

robzienert:
I forgot that we need to support configuring plugin binaries to go to a different bucket. The normal S3 bucket is typically versioned, whereas the plugin binaries bucket should not be. This refactor keeps all existing S3 configuration the same, but adds all new config for the plugin storage.


front50-core/src/main/groovy/com/netflix/spinnaker/front50/plugins/CachingPluginBinaryStorageService.java
```diff
@@ -98,6 +98,7 @@ private synchronized void storeCache(String key, byte[] binary) {
     Path binaryPath = CACHE_PATH.resolve(key);
     if (!binaryPath.toFile().exists()) {
       try {
+        Files.createDirectories(binaryPath.getParent());
```
robzienert: Also squeezing in a wee little bug fix if the cache directory doesn't exist. 🙃

plumpy:
No code reviews? We have a policy about that, you know...

PR #719, opened 2020-02-21
fix(startup): fix startup if a service other than S3 is configured

plumpy:
This reverts #718 and #719.

See #716 for the error; it's the same one.

It seems like you fixed a different problem. The problem I see, however,
is when some _other_ storage service (e.g. Google Cloud Storage) is
configured, but not S3. Your code _always_ binds an `S3StorageService`,
which aborts startup because now there are two `StorageService`
instances bound.

I thought maybe it had something to do with this line in
`S3StorageServiceConfiguration`:

```
@ConditionalOnProperty(value = "spinnaker.s3.storage-service.enabled", matchIfMissing = true)
```

But even if I change it to `matchIfMissing = false`, it still has this
issue, so I didn't dig any deeper.

This reverts commit 5fc94d5 and 5749544.


ezimanyi:
This might be a good motivation to add a Spring Boot startup test to front50; I've added these to a few other services in response to issues like this (ex: spinnaker/igor#410, spinnaker/echo#472, spinnaker/orca#2685).  There could perhaps be a profile to test S3 works and that GCS works, which should give pretty good coverage.

robzienert:
I made an effort to add one yesterday but landed nowhere. I'll take another stab today.

PR #736, opened 2020-03-06
chore(kork): bump korkVersion to 7.25.1

plumpy:


clanesf:
@plumpy @ezimanyi - so it looks like the fix for kork got pulled into the release branch for 1.18.x but the chore here to bump the kork version is targeted for 1.19. Correct me if I'm wrong, but this means we're waiting for 1.19.x for this fix?

plumpy:
This is a different fix. The PRs in question for your fix are spinnaker/igor#50 and spinnaker/clouddriver#4392.

...which I thought were the two services that were using Java 11 in Spinnaker 1.18, but it's actually `front50`, not `clouddriver` :( I'll send a `front50` PR to fix this, sorry.

clanesf:
yeah you're right. Sorry my bad :(

PR #737, opened 2020-03-10
fix(storage): Fix errors on eventual consistency of storage

ezimanyi:
Fixes spinnaker/spinnaker#5557 which was found by the integration tests; this applies the second solution outlined there "Instead of fetching the object from storage after creating it, just return the object we placed in storage.".

* test(storage): Add some tests 

  Add a new test implementation of StorageService that simulates an eventually consistent storage backend, and tests on this new implementation.

  Also add a test on DefaultApplicationDAO to document the current behavior which is to throw an exception when creating an application against an eventually consistent data store.

* refactor(storage): Strengthen contract of StorageService 

  A StorageService should not implement any business logic and should just store the provided item to the backing store as given.

  There are a couple of places where the storage service sets a modified timestamp (in the Azure and SQL backends); arguably this should be done upstream so that this is independent of the backing store but for now we'll leave those there.

  Add javadoc to storeObject to explicitly strengthen the contract of this method, which all current implementations respect. In particular, require that the input object reflect exactly what was stored in the persistent store upon returning from the method. Thus, implmementations can mutatate the input object before storing it, but must ensure that mutation is reflected on the input object.

* refactor(core): Strengthen contract of StorageServiceSupport.update 

  Now that StorageService.storeObject explicitly guarantees that the state of the object upon returning from the function relfects exactly what was stored, we can strengthen the contract of StorageServiceSupport.update.

  In particular, we'll allow update to modify the supplied object (for example to add last modified data) but will require that the state of the input object at the end of update reflect exactly what was stored to the persistent store.

  In order to help enforce this contract, we'll make update final so that implementations cannot override it in a way incompatible with this contract. There were a few implementations that did override update, but these all just did pre-update mutation/validation then called super.update. To handle these cases, expose a preUpdate function where implementations can specify this pre-update mutation/validation.

* fix(storage): Short-circuit findById after update 

  Now that we have strengthened the contract of update, we can short-circuit places where we call update to store an object, then immediately call findById to get it back again. Given this strengthened contract, we are guaranteed that the object in the persistent store is identical to the object we just stored and can avoid an extra call to the persistent store.

  This is particularly important for persistent stores that don't have strong read-after-write consistency (such as s3). In these cases, immediately fetching the the object we just wrote may actually return a prior version of the object (or a 404 if the object was just created). Returning the object we know we just wrote avoids this race condition between our call to fetch the object and the eventual consistency of the data store.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,11 @@ allprojects { project ->
       testImplementation("org.hamcrest:hamcrest-core")
       testRuntimeOnly("cglib:cglib-nodep")
       testRuntimeOnly("org.objenesis:objenesis")
+      testRuntimeOnly "org.junit.vintage:junit-vintage-engine" // Required for Spock tests to execute along with Junit5 tests.
+    }
+
+    test {
+      useJUnitPlatform()
     }
   }
 }
--- a/front50-core/front50-core.gradle
+++ b/front50-core/front50-core.gradle
@@ -32,4 +32,8 @@ dependencies {
   implementation "com.netflix.spinnaker.kork:kork-hystrix"
   implementation "com.netflix.spinnaker.kork:kork-security"
   implementation "com.netflix.spinnaker.kork:kork-artifacts"
+
+  testImplementation "org.assertj:assertj-core"
+  testImplementation "org.junit.jupiter:junit-jupiter-api"
+  testImplementation "org.junit.platform:junit-platform-runner"
 }
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageService.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageService.java
@@ -49,6 +49,31 @@ public interface StorageService {
     }
   }
 
+  /**
+   * Stores the provided object into the persistent store, using the provided key.
+   *
+   * <p>Implementations may modify the object before storing it (such as to add timestamps) but must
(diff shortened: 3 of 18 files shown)


ezimanyi:
@ajordens @robzienert : I'd definitely like someone from Netflix to take a look at this to make sure that this change won't negatively affect the SQL storage implementation. As far as I can tell it won't but want someone who is more familiar with it to take a look.  Thanks!

ajordens:
If memory serves, the `S3StorageService` might be setting the last modified timestamp to some attribute of the s3 object metadata versus just trusting whatever was supplied.

I'll tag @srekapalli since he's been working with `front50` most recently from our side. We're off S3 but worth confirming no impact to the SQL paths.

ezimanyi:
@ajordens : Good catch, we do indeed set the `lastModified` time of of the object to the s3 metadata time in `fetchObject` (and do the same for the GCS back-end).  (That's actually the root cause of the failing test here.)

This is tricky to solve, as the response from a PUT request to s3 doesn't include the `Last-Modified` header, so the only way to check what got written to the metadata is to make a subsequent `HEAD` or `PUT` request...which is exactly the request that is intermittently failing.

I wonder if the better answer here is just not returning the object at all from a `create` request and forcing consumers to fetch it after create if they need it (which I suspect most don't) and let those that do handle retries on a 404.

ezimanyi:
Going to close this for now; will re-open if I have time to implement a better solution for this that doesn't have the bug noted above.

PR #740, opened 2020-03-13
fix(s3): Update aws-sdk to v1.11.717

FaHeymann:
Trying to assume an IAM role with the EKS pod identity webhook results in a stackoverflow error with Spinnaker v1.19.0 (Front 50 v0.22.0). The issue is described in https://github.com/aws/aws-sdk-java/issues/2136 and updating the aws-sdk should fix it.

Edit: I just saw https://github.com/spinnaker/spinnaker/issues/5521 and https://github.com/spinnaker/kork/pull/523
I don't know how the dependency management is supposed to work here, but in the current master the aws-sdk version still resolves to 1.11.415

### DIFF
--- a/front50-s3/front50-s3.gradle
+++ b/front50-s3/front50-s3.gradle
@@ -24,8 +24,8 @@ dependencies {
   implementation "com.netflix.spinnaker.kork:kork-aws"
   implementation "com.netflix.spinnaker.kork:kork-security"
   implementation "com.netflix.spinnaker.kork:kork-hystrix"
-  implementation "com.amazonaws:aws-java-sdk-s3"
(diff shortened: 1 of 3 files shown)


FaHeymann:
I tested with a self-build image now, and with this change it works as expected.

To be able to access the credentials file in the injected volume a security context has to be set.
```
kubernetes:
  serviceAccountName: spinnaker-front50
  securityContext:
    fsGroup: 102
```
in `~.hal/default/service-settings/front50.yml`

clareliguori:
Closing this, so we can fix the root cause in kork (https://github.com/spinnaker/kork/pull/554).  Thank you for debugging this!

PR #755, opened 2020-03-26
Plugin upsert fix

srekapalli:
-- Upsert release nodes in the plugin info w/o replacing the old release information (which we require to rollback to the previous version if needed)

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugininfo/PluginInfoService.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugininfo/PluginInfoService.java
@@ -55,19 +55,14 @@ public class PluginInfoService {
 
     try {
       PluginInfo currentPluginInfo = repository.findById(pluginInfo.getId());
-      List<PluginInfo.Release> newReleases = new ArrayList<>();
+      List<PluginInfo.Release> newReleases = new ArrayList<>(pluginInfo.getReleases());
       List<PluginInfo.Release> oldReleases = new ArrayList<>(currentPluginInfo.getReleases());
-      new ArrayList<>(pluginInfo.getReleases())
+      new ArrayList<>(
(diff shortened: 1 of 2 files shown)


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugininfo/PluginInfoService.java
```diff
@@ -53,7 +54,19 @@ public PluginInfo upsert(@Nonnull PluginInfo pluginInfo) {
...
+      new ArrayList<>(newReleases)
+          .forEach(
+              release -> { // remove it from the old release list if it is being updated.
+                oldReleases.removeIf(
+                    oldRelease -> oldRelease.getVersion().equals(release.getVersion()));
+              });
+
+      List<PluginInfo.Release> allReleases = new ArrayList<>();
+      Stream.of(oldReleases, newReleases).forEach(allReleases::addAll);
+      pluginInfo.setReleases(allReleases);
```
robzienert: I'm not certain we want to allow updating previous releases - in my mind these are immutable: We wouldn't want someone to come along and update an old release to point to a new artifact, for example. Shouldn't the upsert operation just add new releases to the existing releases list? If they're trying to update a previous release, I'd have expected an error. Happy to be wrong, here, though.

cfieber: if it does need to happy to fix a broken publication or something it would be best to be explicit (deleting an old version first, etc)

srekapalli: Yh.. Add the new release to the existing release is what I started with, but I was thinking about other attributes on the release say for eg `required` restrictions etc which I thought we can update. May be..?  Not a usecase to worry about now .. I will just add the new release to the current release .

PR #767, opened 2020-04-10
chore(dependencies): Autobump korkVersion

spinnakerbot:


plumpy:
@robfletcher This is breaking because of [your bump to `kotlinx-coroutines` 1.3.5](https://github.com/spinnaker/kork/pull/558). Could you take a look or nominate someone at Netflix to fix it? Thanks!

Running `./gradlew :front50-sql:test` gets you this:
`NoClassDefFoundError: kotlin/coroutines/AbstractCoroutineContextKey`

It still compiles okay, so I'm guessing there's some other library that has a hard dependency on an old version of coroutines? (And maybe that library is `strikt`?)

robfletcher:
It's breaking because for some reason front50 has the version of Kotlin pinned back to an arbitrary old version. See https://github.com/spinnaker/front50/pull/772

PR #772, opened 2020-04-14
chore(dependencies): Unpin org.jetbrains.kotlin

ezimanyi:
Bumping kork on front50 is failing because at some point recently we upgraded kotlin, and it looks like it has a new artifact that isn't in the pinned version.

I'm not sure why the pin was added initially, but the tests all pass when it is removed. When this is removed, kork can successfully be bumped.

### DIFF
--- a/gradle/kotlin.gradle
+++ b/gradle/kotlin.gradle
@@ -17,16 +17,6 @@
 apply plugin: "nebula.kotlin"
 apply plugin: "kotlin-spring"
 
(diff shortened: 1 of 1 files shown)


ezimanyi:
This should unblock the auto-bump in #767 that is now quite a bit behind the latest kork.

@ajordens : Do you remember why this was pinned initially and if there's a reason we shouldn't un-pin it?

PR #783, opened 2020-04-22
refactor(gcs): Replace complex Closure logic with Callable

ezimanyi:
* refactor(gcs): Update retry to new format 

  This updates the GCS storage backend to consume the new retry implementation that can return typed objects. This commit does the minimal change required to get the front50 GCS implementation to work with the new retry implementation, which is:
  (1) Delete the GcsSafeRetry that extends the now final GoogleCommonSafeRetry and just use a GoogleCommonSafeRetry directly
  (2) Replace the Long values in the config with int values (as nobody is waiting 24 days between retries)
  (3) Instead of storing all these retry values directly on GcsStorageService, just create the GoogleCommonSafeRetry that uses them in the constructor
  (4) Replace the catch blocks that were expecting some mix of exceptions to now catch GoogleApiException and return a GcsProviderOperationException wrapping it with a description of what was happening.

* refactor(gcs): Replace complex Closure logic with Callable 

  Now that the retry handler can accept a callable and can return a typed object, replace the extremely complex closure logic in this file with callables defined as lambdas.

* fix(gcs): Remove extraneous log statement 

  When we don't find an object, we log then throw a NotFoundException. In general the benefit of logging then throwing an exception is questionable; it's better to just include the details on the exception which will either be handled or logged anyway.

  In this case, there are some cases where a NotFoundException is a reasonable outcome, and we catch it and proceed. In those cases, the log statement is just adding noise to the logs. Let's remove it and improve the message on the exception, so that cases that do end up propagating or logging that exception will still get a good message but those that choose to ignore the exception don't add noise to the logs.


ezimanyi:
This is the change to `front50` that motivated my refactor of the retry logic in clouddriver.  Ideally front50 wouldn't depend on clouddriver as a library, but that's a problem for another day.

I've marked this as a draft PR because it won't compile until we bump the clouddriver version; when I release a clouddriver I'll update this PR to pull in the new clouddriver version (as the autobump will fail because it needs these changes).

ezimanyi:
The last commit fixes spinnaker/spinnaker#5698, which is how I ended up here in the first place.

ezimanyi:
I cut a new version of clouddriver, and updated this PR to bump clouddriver, so it's no longer a draft.  (The only change is the clouddriver bump and merging in master.)

maggieneterval:
👍 👍 👍 👍 👍

PR #786, opened 2020-04-24
refactor(core): clean up halconfig/front50.yml

maggieneterval:
* refactor(core): add README to halconfig directory 

* refactor(core): remove cassandra config from halconfig/front50.yml 

  Cassandra was [removed](https://github.com/spinnaker/front50/pull/353) as a storage provider in 2018, so we should no longer need this config block disabling it.

### DIFF
--- a/halconfig/front50.yml
+++ b/halconfig/front50.yml
@@ -2,9 +2,6 @@ server:
   port: ${services.front50.port:8080}
   address: ${services.front50.host:localhost}
 
-cassandra:
-  enabled: false
-
 hystrix:
   command:
     default.execution.isolation.thread.timeoutInMilliseconds: 15000
--- a/halconfig/front50.yml
+++ b/halconfig/front50.yml
@@ -1,14 +1,3 @@
 server:
   port: ${services.front50.port:8080}
   address: ${services.front50.host:localhost}
-
-hystrix:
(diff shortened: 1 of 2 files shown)


halconfig/front50.yml
```diff
@@ -2,9 +2,6 @@ server:
   port: ${services.front50.port:8080}
   address: ${services.front50.host:localhost}
 
-cassandra:
-  enabled: false
-
 hystrix:
```
maggieneterval, line 5: @ezimanyi I was unsure whether this `hystrix` config block is necessary: it doesn't look like any of the other microservices override any hystrix settings in their config, and there is no corresponding `hystrix` block in the non-Halyard front50-web config. I left it in just to be safe, but if non-Halyard users have been fine without these Halyard-specific `hystrix` overrides then my thought was that kleat-deployed Spinnakers shouldn't need them either?

ezimanyi: Yeah, I suspect these are not necessary, and it's not clear how much though was put into tuning these numbers.  I'd say let's take it out and then if we do find we need this, add back in sensible values in the base config in `front50-web`.

maggieneterval: Sounds like a plan; I've removed it!

PR #791, opened 2020-04-28
feat(plugins): Plugin version pinning

robzienert:
Adds new metadata store for pinning plugin versions to a server group. This
ensures a server group will continue to load the same plugins even through
instance replacement on long-lived server groups.

Metadata is regularly cleaned up, leaving the last N records for a particular
cluster, by default leaving the last 10 (which is likely excessive).

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginVersionController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginVersionController.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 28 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginVersionController.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.netflix.spinnaker.front50.model.plugins.PluginInfo;
+import com.netflix.spinnaker.front50.model.plugins.PluginVersionPinningService;
+import java.util.Map;
+import lombok.Value;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping("/pluginVersions")
+public class PluginVersionController {
```
jonsie, line 26: Do we need an endpoint for defining the "latest" versions for a service?  I.E., in the case wherein the server group is unknown (because it has not been deployed yet) and we want to define the versions for that next server group.

robzienert: I don't think that would be an endpoint that falls in scope of this PR. We'll need ways to actuate the versions, but there's a lot more than just that endpoint involved in making the feature work - I'd rather tackle that when we get to the overall feature. WDYT?

jonsie: That works for me. 👍

jonsie:
I think we'll want to add a `DefaultTableDefinition` and a migration to create the `plugin_versions` SQL table.

PR #797, opened 2020-04-29
chore(gcs): Add deafult rootFolder name

ezimanyi:
This was being defaulted by Halyard, but will no longer be defaulted by kleat; add this default to front50 so that users using kleat get the same default as Halyard users. This doesn't affect users who explicitly specify it (or Halyard users as Halyard always explicitly defaults the field).


ezimanyi:
I guess it would have been easier if I didn't do these one-by-one... I think I missed this one as we actually wrote the GCS proto back on January 16th in the initial commit!

PR #804, opened 2020-05-06
feat(plugins): Specify preferred plugin info release

jonsie:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugins/PluginInfoService.java
```diff
@@ -103,6 +104,28 @@ public PluginInfo deleteRelease(@Nonnull String id, @Nonnull String releaseVersi
...
+      if (preferred) {
+        pluginInfo.getReleases().stream()
+            .filter(it -> !it.getVersion().equals(release.get().getVersion()))
+            .forEach(it -> it.setPreferred(false));
+      }
+
+      repository.update(pluginInfo.getId(), pluginInfo);
+    }
+
+    return release.orElse(null);
```
robzienert, line 126: nit: `release.map(r -> { /* do your stuff */).orElse(null)` -- then you can avoid the `isPresent()` and explicit `get()` call.

PR #808, opened 2020-05-06
fix(plugins): Authorize destructive plugin mgmt endpoints

robzienert:
Danger danger.

We don't want to authz the create endpoints, because anyone can publish a plugin and that's a-ok.

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginInfoController.java
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginInfoController.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.front50.model.plugins.PluginInfoService;
 import java.util.Collection;
 import java.util.Optional;
 import org.springframework.http.HttpStatus;
+import org.springframework.security.access.prepost.PreAuthorize;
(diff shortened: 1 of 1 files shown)


front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PluginInfoController.java
```diff
@@ -56,6 +62,7 @@ PluginInfo upsert(@RequestBody PluginInfo pluginInfo) {
     return pluginInfoService.upsert(pluginInfo);
   }
 
+  @PreAuthorize("@pluginInfoController.admin")
```
cfieber, line 65: I think you should be able to `@PreAuthorize("@fiatPermissionEvaluator.isAdmin()")` (we are doing that elsewhere)

robzienert: In this case, I don't need to inject it into the controller then, right?

cfieber: nope, it should be all wired up for security annotation processing

PR #811, opened 2020-05-07
fix(plugins): Delete binary when release info is deleted

robzienert:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugins/PluginInfoService.java
```diff
@@ -35,10 +36,15 @@
 public class PluginInfoService {
 
   private final PluginInfoRepository repository;
+  private final PluginBinaryStorageService storageService;
```
cfieber, line 39: this thing is optional in PluginBinaryController, does it need to be optional here (or alternatively not-optional over there?)

robzienert: You right. You right.

PR #826, opened 2020-05-19
fix(plugins): Set preferred release to false when upserting plugin info; fix semver validation on release version

jonsie:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugins/PluginInfo.java
```diff
@@ -77,7 +77,7 @@ public void setReleaseByVersion(String version, Release release) {
   /** A singular {@code PluginInfo} release. */
   @Data
   public static class Release {
-    public static final String VERSION_PATTERN = "^[1-9]\\d*\\.\\d+\\.\\d+(?:-[a-zA-Z0-9]+)?$";
+    public static final String VERSION_PATTERN = "^[0-9]\\d*\\.\\d+\\.\\d+(?:-[a-zA-Z0-9]+)?$";
```
jonsie, line 80: Versions can start with 0, not sure why I didn't catch this.

christopherthielen: any reason this regex isn't just `^\\d+\\.\\d+\\.\\d+(?:-[a-zA-Z0-9]+)?$";` ?

PR #828, opened 2020-05-20
perf(sql): Support for bulk sql imports

ajordens:
This PR also adjusts the bulk import behavior for entity
tags to avoid a one-by-one fetch after the import succeeds.

Given our use-cases, I believe this unnecessarily adds to the
import time.

### DIFF
--- a/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -244,6 +244,20 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
     User authenticatedUser = new User();
     authenticatedUser.setUsername(AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"));
 
+    if (service instanceof BulkStorageService) {
+      String lastModifiedBy = AuthenticatedRequest.getSpinnakerUser().orElse("anonymous");
+      Long lastModified = System.currentTimeMillis();
+
+      items.forEach(
+          item -> {
+            item.setLastModifiedBy(lastModifiedBy);
+            item.setLastModified(lastModified);
+          });
+
+      ((BulkStorageService) service).storeObjects(objectType, items);
+      return;
+    }
+
     Observable.from(items)
         .buffer(10)
         .flatMap(
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
@@ -60,7 +60,7 @@ class SqlStorageService(
   private val sqlRetryProperties: SqlRetryProperties,
   private val chunkSize: Int,
   private val poolName: String
-) : StorageService, AdminOperations {
+) : StorageService, BulkStorageService, AdminOperations {
 
   companion object {
     private val log = LoggerFactory.getLogger(SqlStorageService::class.java)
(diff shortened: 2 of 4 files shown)


front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -165,6 +165,52 @@ class SqlStorageService(
     }
   }
 
+  override fun <T : Timestamped?> storeObjects(objectType: ObjectType, allItems: Collection<T>) {
+    allItems.chunked(chunkSize).forEach { items ->
+      try {
+        jooq.transactional(sqlRetryProperties.transactions) { ctx ->
+          ctx.batch(
```
ajordens, line 172: Started this as a proof-of-concept so didn't include the `} catch (e: SQLDialectNotSupportedException) {` handling that we're doing in the singular `storeObjects()`.

Assuming we need that (we still do right?) than my thinking would be to flip back to the same look up individual record -> update or insert individual record pattern. It would certainly be slower.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/EntityTagsController.groovy
```diff
@@ -87,7 +87,7 @@ class EntityTagsController {
     }
 
     taggedEntityDAO.bulkImport(tags)
-    return findAllByIds(tags.findResults { it.id })
```
ajordens, line 90: This is actually a fairly significant contributor to overall slowness.

The entity tag objects themselves aren't changing aside from `lastModified` and `lastModifiedBy` which aren't a deal breaker. As it stands the object references are being updated directly so they're going to be picked up here anyhow.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -244,6 +244,17 @@ public void bulkImport(Collection<T> items) {
     User authenticatedUser = new User();
     authenticatedUser.setUsername(AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"));
 
+    if (service instanceof BulkStorageService) {
+      items.forEach(
+          item -> {
+            item.setLastModifiedBy(AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"));
+            item.setLastModified(System.currentTimeMillis());
+          });
```
marchello2000, line 252: prob should take out the values being set to outside the loop, faster and then update time will be same for all items
```suggestion
      long updateTime = System.currentTimeMillis();
      String modifiedBy = AuthenticatedRequest.getSpinnakerUser().orElse("anonymous");
      
      items.forEach(
          item -> {
            item.setLastModifiedBy(modifiedBy);
            item.setLastModified(updateTime);
          });
```

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/EntityTagsController.groovy
marchello2000: yeah, this makes sense!

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
ajordens: Gotcha, good call ... I'll make this change.

ajordens:
Add some record-by-record fallbacks to handle the `SQLDialectNotSupportedException`.

Thoughts @marchello2000 ?

marchello2000:
love it!, thank you!

My only concern is that the `chunkSize` is pretty big - 1k - (https://github.com/spinnaker/front50/blob/master/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt#L52)

which might mean we blow packet size limit. The only other place we use the `chunkSize` is in a select where it's negligible as far as the packet size... 
current MySQL version has packet size of 4mb, which means each object should be under 4kb so it's probably fine

ajordens:
tbh I initially had chunkSize at 100 before I noticed we actually had a variable for it ... I've no qualms about pulling it back.

PR #830, opened 2020-05-20
fix(fiat): only attempt to sync permissions when fiat is enabled

maggieneterval:
In [this PR](https://github.com/spinnaker/front50/pull/619), we added logic to sync roles after an application is created. Let's only attempt to sync roles if the Fiat client is actually enabled. Elsewhere in Front50, such as in ApplicationPermissionsService.syncUsers, we only attempt to call Fiat to sync these roles if FiatClientConfigurationProperties is enabled, if FiatService is present, and if FiatConfigurationProperties.roleSync is enabled. Let's mirror this logic in the ApplicationsController.

cc @AbdulRahmanAlHamali

### DIFF
--- a/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
+++ b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
@@ -1,7 +1,9 @@
 package com.netflix.spinnaker.front50.controllers.v2
 
+import com.netflix.spinnaker.fiat.shared.FiatClientConfigurationProperties
 import com.netflix.spinnaker.fiat.shared.FiatService
 import com.netflix.spinnaker.front50.ServiceAccountsService
+import com.netflix.spinnaker.front50.config.FiatConfigurationProperties
 import com.netflix.spinnaker.front50.controllers.exception.InvalidApplicationRequestException
 import com.netflix.spinnaker.front50.events.ApplicationEventListener
 import com.netflix.spinnaker.front50.exception.NotFoundException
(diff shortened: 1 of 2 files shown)


maggieneterval:
For context, the primary goal of this PR is to clean up our end-to-end tests' Front50 logs, which are currently littered with these errors, in order to debug some actual issues.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ApplicationsController.groovy
```diff
@@ -107,10 +115,12 @@ public class ApplicationsController {
   @RequestMapping(method = RequestMethod.POST)
   Application create(@RequestBody final Application app) {
     Application createdApplication = getApplication().initialize(app).withName(app.getName()).save()
-    try {
-      fiatService.ifPresent { it.sync() }
-    } catch (Exception ignored) {
-      log.warn("failed to trigger fiat permission sync", ignored)
+    if (fiatClientConfigurationProperties.isEnabled() && fiatConfigurationProperties.getRoleSync().isEnabled() && fiatService.isPresent()) {
```
cfieber, line 118: I think you want to wire in the FiatStatus component here instead and check isEnabled on that, that is used elsewhere as the determiner that fiat is enabled (it is just a wrapper around FiatClientConfigurationProperties that supports toggling the enabled status at runtime)

maggieneterval: Thank you, sounds good! Yeah, was not sure if there was a reason we were using `FiatStatus` in Gate but not Front50, but am happy to wire that in to this service (can also replace the existing checks with that in a separate commit for consistency, as I think there are only a handful of places to update). Will ping for a final look once that is complete. Thanks again for pointing that out! 🙏

maggieneterval: Added two additional commits: one to wire a `FiatStatus` bean, and one to consume it in the `ApplicationsController`. I didn't end up refactoring the other consumers of `FiatClientConfigurationProperties.isEnabled` because we might want to end up cherry-picking this to help debug the release branch tests as well, and wanted to keep this fix as minimal as possible.

PR #835, opened 2020-05-22
fix(plugins): Store plugin info release date as ISO-8601 instant

jonsie:


front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/plugins/DefaultPluginInfoRepository.java
```diff
@@ -71,4 +74,23 @@ public PluginInfo create(String id, PluginInfo item) {
...
+  public void update(String id, PluginInfo item) {
+    item.getReleases().forEach(release -> release.setDate(releaseDateInstant(release.getDate())));
+    super.update(id, item);
+  }
+
+  // The plugin info model is defined in various places, and the release date field can be
+  // serialized as either a unix timestamp or an ISO-8601 instant.  We store it as an instant.
+  private String releaseDateInstant(String releaseDate) {
+    try {
+      int timestamp = Integer.parseInt(releaseDate);
```
cfieber, line 88: think this should be Long.parseLong for a timestamp

PR #836, opened 2020-05-27
refactor(core): Move logic out of Application into ApplicationService

robzienert:
- Converts `Application` to a POJO: Moves all existing logic into
  `ApplicationService`
- Simplifies persistence logic to remove concept of "current" and "updated"
  states that developers need to reason about. Unnecessary complexity.
- Changed `ApplicationEventListener` interface to extend `Consumer`. No
  longer supports `rollback`, which was unused anyway; improved consumer
  interface to use `ApplicationModelEvent` to now include `Type` when
  processing the event.

This is a breaking change due to the refactors of `ApplicationEventListener`. I can roll this change back if people would like, but it was bothering me. 😬 

This seemed like a big enough change on its own to submit. Once this gets merged, then there's only a handful of other gnarly classes to convert to Java.


jervi:
I _think_ this PR made it impossible to run front50 with multiple storage providers enabled simultaneously.

For instance with SQL and S3 enabled:

```
Parameter 0 of method applicationDAO in com.netflix.spinnaker.front50.config.CommonStorageServiceDAOConfig required a single bean, but 2 were found:
	- s3StorageService: defined by method 's3StorageService' in class path resource [com/netflix/spinnaker/front50/config/S3Config.class]
	- sqlStorageService: defined by method 'sqlStorageService' in class path resource [com/netflix/spinnaker/config/SqlConfiguration.class]
```

PR #837, opened 2020-05-27
refactor(core): Convert remaining Groovy to Java

robzienert:
Aside from https://github.com/spinnaker/front50/pull/836, this PR represents the remaining Groovy left in the `main` source set for `front50-core`. This does not have a dependency on #836 or vice-versa.

While the PR is large, this is just a direct migration from Groovy to Java. The large delta is due to porting heavy use of Groovy-only code to Java. One new class exists, `UntypedUtils`, which was written to help with a bunch of the logic around `Application` and friends that depended heavily on metaclassing. This PR doesn't fix the anti-pattern, it only makes it possible in vanilla Java via reflection.

I don't have a good solution as yet of how we could wrangle `Application` into something a little more sane. I think namespacing like configs into their own buckets would be an improvement, but I have no idea what a migration to that would look like without some dramatic ripple effects.


front50-core/src/main/groovy/com/netflix/spinnaker/front50/UntypedUtils.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public static boolean hasProperty(Object obj, String propertyName) {
+    return Arrays.stream(obj.getClass().getDeclaredFields())
+        .anyMatch(it -> it.getName().equals(propertyName));
+  }
+
+  @SneakyThrows
+  public static Map<String, String> getProperties(Object obj) {
+    Map<String, String> m = new HashMap<>();
+    Arrays.stream(obj.getClass().getDeclaredFields())
```
cfieber, line 55: this won't pick up anything from a superclass or interface (not sure that matters for how we are using it, and I can't imagine we will write any new code using this class)

front50-core/src/main/groovy/com/netflix/spinnaker/front50/UntypedUtils.java
```diff
@@ -0,0 +1,75 @@
...
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+import lombok.SneakyThrows;
+
+/** Helper class for dealing with untyped data / Groovy migration. */
+@NonnullByDefault
+public class UntypedUtils {
```
cfieber, line 30: maybe worth sticking an `@Deprecated` on here with a note that this is only here to support groovy -> java

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public void setId(String id) {
+    super.put("id", id);
+  }
+
+  @Override
+  @JsonIgnore
+  public Long getLastModified() {
+    String updateTs = DefaultGroovyMethods.asType(super.get("updateTs"), String.class);
```
cfieber, line 44: Should this be an untypedUtils thing (or something not DefaultGroovyMethods anyways)

robzienert:
@ezimanyi 

> A bunch of the new classes have getters/setters (replacing the Groovy auto-generated ones) that could probably be Lombok to remove some boilerplate.

I don't trust Lombok when it comes to serialization. I realize lombok will make code more terse, but I never trust that a new Lombok annotation won't ruin Jackson serialization. I feel like Lombok'ing existing classes can come as a separate change. wdyt?

front50-core/src/main/groovy/com/netflix/spinnaker/front50/model/pipeline/Pipeline.java
robzienert: Oops. Missed this conversion. Thanks!

front50-core/src/main/groovy/com/netflix/spinnaker/front50/UntypedUtils.java
robzienert: Good call, will do.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/UntypedUtils.java
robzienert: I don't know if we need the additional complexity. Hopefully we can land on something to get rid of this class in the medium-future, but as you say, I don't think we'll be extending this class any time soon.

ezimanyi:
@robzienert : I'm fine with no Lombok; I've also had issues where Lombok was more trouble than it was worth (and also think that at times it makes it too easy to make everything public).  I think we've talked about even moving from Lombok to something else but decided there were enough other things (like removing Groovy) we should do first.  So yeah, totally fine with this as is, and even if we don't later lombok-it.

PR #839, opened 2020-05-27
fix(sql): `/ServiceAccounts` fails when sql is only active provider

jervi:
This was a bit of a weird error. After we turned off the S3 provider in Front50, the `/serviceAccounts` endpoint stopped working, failing with the following error message: "Configured storage service does not support service account permissions".
This happens when there's no `ServiceAccountsService` bean (which it really wasn't). This bean is autoconfigured using `@ConditionalOnBean(ServiceAccountDAO.class)`, and according to Spring there's no ServiceAccountDAO in the context. But when inspecting the context, it actually _is_ a `ServiceAccountDAO` present. I think that for some reason the `ServiceAccountDAO` bean is created _after_ Spring tries to wire up `ServiceAccountsService`, but this doesn't happen with the other providers. I have no idea why. The only thing I can think of is that `SqlConfiguration.kt` is the first provider config written in Kotlin, but I don't really understand why that should matter. I'd appreciate any insight into the matter.

I have fixed the issue by creating a new `@ConditionalOnAnyProviderExceptRedisIsEnabled` annotation that returns true if any of the providers except Redis are enabled. I've tested running the app with only SQL enabled, only Redis enabled*, and both SQL and Redis enabled. All works now.

\* This didn't work before, as @robzienert correctly theorised.


jervi:
front50-local.yml:
```
spinnaker:
  migration:
    enabled: false
  s3:
    enabled: false
  redis:
    enabled: false
  oracle:
    enabled: false
  gcs:
    enabled: false
  azs:
    enabled: false
sql:
  enabled: true
  connectionPools:
    default:
      default: true
      jdbcUrl: jdbc:mysql://localhost:3306/front50
      user: root
  migration:
    user: root
    jdbcUrl: jdbc:mysql://localhost:3306/front50
```

jervi:
@adriengentil @robzienert PTAL

adriengentil:
I played a little in the afternoon and I managed to make it work by just adding the `@Configuration` annotation to the class `CommonStorageServiceDAOConfig`. I suppose that in that way, Spring can now resolve automatically the dependencies between the beans which was not the case before.

I don't know if there are any downsides to this solution and if there is a reason that explains why it was not there before.

robzienert:
@adriengentil is on the right path.

I don't know why the config was written the way it was: Adding `@Configuration` to `CommonStorageServiceDAOConfig` will fix it up. While you're at it, making `SqlConfiguration` not extend from `CommonStorageServiceDAOConfig` would be a good idea.

jervi:
But what about all the other providers (except Redis)? They also extend `CommonStorageServiceDAOConfig`.

robzienert:
Oh, I see. Heh, I get it now. Give me a few minutes to dig in a little more.

robzienert:
So, here's the thing (being overly explanatory, repeating some of what you said, for other on-lookers):

**Skip if you know front50:** The Redis provider, for better _and_ worse, has been overlooked while new features have been developed. Each object type (applications, pipeline configs, service accounts, etc.) has it's own DAO that's written, and for each persistence backend, there is a DAO class written for it. S3, SQL and so-on all extend from a common base implementation so once you've written the DAO once, you've written it for everything. Redis, on the other hand, doesn't extend from from this common implementation and requires implementation on its own for support.

The following object types do not have Redis DAOs written for them:

1. Application Permissions
1. Service Accounts
1. Snapshots (although this is for really purposed for blobstores only, I think?)
1. Entity Tags
1. Delivery Configs
1. Plugin Info
1. Plugin Versions

This is a pretty long list. If someone were actually running Front50 on Redis today, they would have a very crippled experience, if it even works at all for many use cases.

The proposed solution in this PR would break the Redis implementation because no `ServiceAccountDAO` exists for the Redis backend. On the other hand, it's my opinion that Front50 is so crippled using the Redis backend that Spinnaker is basically broken anyway.
That said, I don't think we can merge this PR as-is because there "might" be people using it, so we need to go through a deprecation cycle to allow people time to move off.

What I'd suggest is keeping what you have here, but also adding a new `RedisServiceAccountDAO`. After that, I'd feel reasonably assured that we could merge this PR.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/config/CommonServiceConfig.java
```diff
@@ -1,47 +0,0 @@
...
-import java.util.Optional;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
-
-@Configuration
-public class CommonServiceConfig {
-
-  @Bean
-  @ConditionalOnBean(ServiceAccountDAO.class)
```
cfieber, line 33: `@ConditionalOnBean` stuff is pretty finicky outside the AutoConfiguration lifecycle and we are relying on `@ComponentScan` to load all the providers.

I could get behind switching this to `@ConditionalOnExpression("not ${spinnaker.redis.enabled:false}")` (that SpEL is likely maybe wrong)

cfieber: I guess this would mess up someone who is trying to have redis and another provider enabled for migration buutttt I'm still okay with that personally

jervi:
Thank you for the great feedback @robzienert and @cfieber. I've updated the PR and I think this solves all the issues.

plumpy:
@Mergifyio backport release-1.20.x

plumpy:
@jervi do you think you could send a PR to merge this back to release-1.20.x? Thanks.

jervi:
@plumpy https://github.com/spinnaker/front50/pull/848

PR #841, opened 2020-05-28
refactor(web): Convert web module to Java

robzienert:
When you're on a roll, you're on a roll.

This is the entire module.


front50-core/src/main/groovy/com/netflix/spinnaker/front50/ServiceAccountsService.java
```diff
@@ -58,11 +58,11 @@ public ServiceAccountsService(
     this.fiatPermissionEvaluator = fiatPermissionEvaluator;
   }
 
-  Collection<ServiceAccount> getAllServiceAccounts() {
+  public Collection<ServiceAccount> getAllServiceAccounts() {
     return serviceAccountDAO.all();
   }
 
-  ServiceAccount createServiceAccount(ServiceAccount serviceAccount) {
+  public ServiceAccount createServiceAccount(ServiceAccount serviceAccount) {
```
robzienert, line 65: These two methods were being accessed despite being out of scope.

front50-web/front50-web.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
   implementation "org.springframework.boot:spring-boot-starter-actuator"
   implementation "com.netflix.spinnaker.fiat:fiat-api:$fiatVersion"
   implementation "com.netflix.spinnaker.fiat:fiat-core:$fiatVersion"
+  implementation "com.netflix.spinnaker.kork:kork-artifacts"
```
robzienert, line 39: Groovy was referencing artifacts it didn't have access to.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/NotificationController.java
```diff
@@ -0,0 +1,113 @@
...
+      NotificationDAO.NOTIFICATION_FORMATS.forEach(
+          it -> {
+            if (UntypedUtils.hasProperty(global, it)) {
+              if (!UntypedUtils.hasProperty(notification, it)) {
+                UntypedUtils.setProperty(notification, it, new ArrayList<>());
+              }
+              UntypedUtils.setProperty(notification, it, UntypedUtils.getProperty(global, it));
+            }
+          });
+    }
```
robzienert, line 67: @ajordens Here's the global notifications. Maybe we can delete?

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.java
```diff
@@ -0,0 +1,244 @@
...
+                                    new FuzzySearch(project, it.getKey(), it.getValue()))))
+            .collect(Collectors.toSet()));
+
+    return new TreeSet<>(items)
+        .stream()
+            .sorted(
+                (a, b) ->
+                    SearchUtils.score(b, normalizedAttributes)
+                        - SearchUtils.score(a, normalizedAttributes))
+            .collect(Collectors.toCollection(LinkedHashSet::new));
```
robzienert, line 187: I broke this monstrosity up into a few blocks and methods. I'm happy to break it up further.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/NotificationController.java
```diff
@@ -0,0 +1,113 @@
...
+    if (level.equals(HierarchicalLevel.APPLICATION)) {
+      final Object global = getGlobal();
+
+      NotificationDAO.NOTIFICATION_FORMATS.forEach(
+          it -> {
+            if (UntypedUtils.hasProperty(global, it)) {
+              if (!UntypedUtils.hasProperty(notification, it)) {
+                UntypedUtils.setProperty(notification, it, new ArrayList<>());
+              }
+              UntypedUtils.setProperty(notification, it, UntypedUtils.getProperty(global, it));
```
cfieber, line 64: I think this needs to be:
```suggestion
              ((List<?>) UntypedUtils.getProperty(notification, it)).addAll(UntypedUtils.getProperty(global, it)));
```

Otherwise it will replace the list instead of adding to it

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -0,0 +1,287 @@
...
+
+    pipelines.sort(
+        (p1, p2) -> {
+          if (p1.getIndex() != null && p2.getIndex() == null) {
+            return -1;
+          }
+          if (p1.getIndex() == null && p2.getIndex() != null) {
+            return 1;
+          }
+          if (!p1.getIndex().equals(p2.getIndex())) {
```
ezimanyi: I think this introduces a new NPE if both indexes are null.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -0,0 +1,287 @@
...
+    return pipelineDAO.all(refresh);
+  }
+
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
+  @RequestMapping(value = "{application:.+}", method = RequestMethod.GET)
+  public List<Pipeline> listByApplication(
+      @PathVariable(value = "application") String application,
+      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh) {
+    List<Pipeline> pipelines =
+        (List<Pipeline>) pipelineDAO.getPipelinesByApplication(application, refresh);
```
ezimanyi: Is this cast safe?  If all implementations really do return this as a `List` should we change the type on the interface?  If not should we create a list from the collection here?

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.java
cfieber: yikes... but yeah it looks like you captured the original essencé d'groovy - nice job

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/NotificationController.java
robzienert: Oops. Right you are.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.java
```diff
@@ -0,0 +1,244 @@
...
+  @ApiOperation(
+      value = "",
+      notes =
+          "Fetch all projects.\n\n    Support filtering by one or more attributes:\n    - ?name=projectName\n    - ?email=my@email.com")
+  @RequestMapping(method = RequestMethod.GET)
+  public Set<Project> projects(
+      @RequestParam(value = "pageSize", required = false) Integer pageSize,
+      @RequestParam Map<String, String> params) {
+    params.remove("pageSize");
+    Set<Project> projects =
```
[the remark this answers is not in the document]
robzienert: IntelliJ didn't complain about it so I figured it would be OK, but I've wrapped `projectDAO.all()` into `new HashSet<>(projectDAO.all())` to be safe. Since `all()` is defined by `ItemDAO<T>`, I'm not comfortable in changing the signature of the method without affecting other places.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.java
```diff
@@ -0,0 +1,244 @@
...
+            .filter(
+                project ->
+                    normalizedAttributes.entrySet().stream()
+                        .anyMatch(
+                            it ->
+                                FUZZY_SEARCH_PREDICATE.test(
+                                    new FuzzySearch(project, it.getKey(), it.getValue()))))
+            .collect(Collectors.toSet()));
+
+    return new TreeSet<>(items)
```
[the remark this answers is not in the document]
robzienert: The signature was a `Set`, although I don't know why. The code was doing sorting on it, so I needed some implementation of `SortedSet`, which landed me on `TreeSet` - maybe this is the wrong implementation.

Alternatively I could change all of this to a `List` and just `distinct` on the stream before doing the sort? Kinda like this instead.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.java
robzienert: Ha. Right you are. I guess this NPE has existed since... 2014. Will fix, good catch.

robzienert:
@cfieber Taking a pass to remove `compileGroovy` is on the agenda once all this is merged!

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.java
robzienert: I'm open to changing the interface to a list, but I think I'll try evaluating / taking that on as part of a separate PR.

front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/v2/ProjectsController.java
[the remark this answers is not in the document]
ezimanyi: Oh I totally understand the result being a `LinkedHashSet` as we do seem to need it sorted...though I kind of like your idea better of making it a `List` with `distinct`.

I meant more that I think the following would be equivalent:
```
items.stream().sort()
```
```
new TreeSet<>(items).stream().sort()
```
so I wasn't sure it was wrapped right before calling `.stream()`.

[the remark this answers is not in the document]
robzienert: Oh boy, yeah ok, now I see what you're saying I missed the `LinkedHashSet` 🤦 .

This is now a List.

clanesf:
Thank you for the fix in Application create to actually check if fiat is enabled before attempting to make the role/sync call! Yet another reason to upgrade to 1.21!

PR #848, opened 2020-05-29
fix(sql): `/ServiceAccounts` fails when sql is only active provider (#839)

jervi:
Cherry pick of 225688b

This was a bit of a weird error. After we turned off the S3 provider in Front50, the `/serviceAccounts` endpoint stopped working, failing with the following error message: "Configured storage service does not support service account permissions".
This happens when there's no `ServiceAccountsService` bean (which it really wasn't). This bean is autoconfigured using `@ConditionalOnBean(ServiceAccountDAO.class)`, and according to Spring there's no ServiceAccountDAO in the context. But when inspecting the context, it actually _is_ a `ServiceAccountDAO` present. I think that for some reason the `ServiceAccountDAO` bean is created _after_ Spring tries to wire up `ServiceAccountsService`, but this doesn't happen with the other providers. I have no idea why. The only thing I can think of is that `SqlConfiguration.kt` is the first provider config written in Kotlin, but I don't really understand why that should matter. I'd appreciate any insight into the matter.

I have fixed the issue by creating a new `@ConditionalOnAnyProviderExceptRedisIsEnabled` annotation that returns true if any of the providers except Redis are enabled. I've tested running the app with only SQL enabled, only Redis enabled*, and both SQL and Redis enabled. All works now.

\* This didn't work before, as @robzienert correctly theorised.

* Add back @ConditionalOnBean and fix visibility of two functions

+ some minor Java style fixes

* Added new conditional annotation and applied it in relevant places

The new @ConditionalOnAnyProviderExceptRedisIsEnabled will be true if any of the providers except Redis are enabled. Whether Redis is enabled or not does not matter.

* Conditionally wire up ServiceAccountsController

Co-authored-by: Cameron Fieber <cfieber@netflix.com>


ezimanyi:
I think that instead of merging this, we should strongly consider rolling back #803.  That was labeled as a fix and cherry-picked into 1.19 despite being a pretty large change, and it looks like it is the cause of this regression here.

I'm surprised that #803 was cherry picked in the first place as it seemed like a reasonably large and risky change (which this bug supports), and am not overly excited about continuing to patch/fix bugs against it in the 1.19 branch instead of just rolling it back.

jervi:
Rolling back #803 in the 1.20 branch sounds good to me. @adriengentil WDYT?

plumpy:
While I agree that #803 never should have been cherry-picked, it went out with the 1.20.1 release. So now it's been in four releases. Rolling it back to me feels potentially as worrisome as merging this... I think we're kinda stuck in a hard situation now and I don't know what the right move is...

front50-core/src/main/groovy/com/netflix/spinnaker/front50/config/annotations/AnyProviderExceptRedisCondition.java
```diff
@@ -0,0 +1,42 @@
...
+        || serviceEnabled("gcs", env)
+        || serviceEnabled("azs", env)
+        || serviceEnabled("swift", env);
+  }
+
+  private static boolean serviceEnabled(String s, Environment env) {
+    return env.getProperty(
+        (s.equals("sql") ? "" : "spinnaker.") + s + ".enabled", Boolean.class, false);
+  }
+}
```
robzienert, line 42: Clever, I like this a lot. Probably a pattern that should be used elsewhere for our crazy conditional configs (in lieu of making config simpler).

ethanfrogers:
I agree, #803 shouldn't have been merged. I was misguided by the fix label (shame on me and should have dug in more thoroughly. 

Like @plumpy said, I think we're in in between two not so great solutions. I think the right thing to do is to merge this and "fix the fix" rather than rolling back #803, however.

front50-core/src/main/groovy/com/netflix/spinnaker/front50/config/annotations/AnyProviderExceptRedisCondition.java
ezimanyi: I'm actually of the opposite opinion about this...splitting up the strings and joining them in code makes it essentially impossible to search for where config properties are used.  Obviously such a search was never perfect before, but this makes it much much harder to find cases where we depend on `redis.enabled` or `sql.enabled`.

I think at the expense of a bit of repetition it would be much clearer to just pass the full property we're checking into this function rather than join strings like this.

Of course ideally we'd map these onto strongly-typed `@ConfigurationProperties` classes which should probably be the direction we go for anything doing significant work on property classes.

cfieber: This could be rewritten as an AnyNestedCondition:
```
public AnyProviderExceptRedisCondition extends AnyNestedCondition {
  AnyProviderExceptRedisCondition() { super(ConfigurationPhase.PARSE_CONFIGURATION) }

  @ConditionalOnProperty("sql.enabled")
  static class SqlEnabled {}
  
  @ConditionalOnProperty("spinnaker.s3.enabled")
  static class S3Enabled {}

  //... and so on
}
```
Which gives the niceness of properly referencing the properties we care about via `@ConditionalOnProperty` and also the niceness of not having insane-o SpEL conditions

jervi: Yeah, that looks better. But it should probably been a part of the original PR, not this cherry pick/backport into the 1.20 branch... I suggest not doing anything here, and rather create a new PR to make the annotation cleaner.

robzienert: I think `AnyNestedCondition` is a much better path... and this PR should probably be updated to use it, seeing how this PR is introducing the annotation to begin with, I don't know why we'd accept this knowing we want to refactor it.

jervi: Sorry, I’m not following you here. The annotation was introduced in #839 and is merged to master. This is a backport of that PR.
Do you mean to create a new PR that refractors the annotation, and when that is merged to master, include that as well in this PR?

robzienert: Oh sorry, I missed that this is a cherry pick. nvm

jervi: https://github.com/spinnaker/front50/pull/858

robzienert:
Merging. We can’t keep the release broken.

jervi:
I don't know why the branch protection is still looking for a Travis status, but I think mergify needs a little help here...

ezimanyi:
I believe it's waiting for approval by a release manager; since @ethanfrogers is release manager this cycle, deferring to him on approval/merge.

cdobbyn:
It looks like it's waiting for this:
```
status-success=continuous-integration/travis-ci/pr
```

based on this:
https://github.com/spinnaker/front50/pull/848/checks?check_run_id=752800211

ethanfrogers:
I merged it with admin privileges. I am not/wasn't sure how to adjust Mergify's rules to exclude Travis from the requirements.

titirigaiulian:
hey, I guess this would be part of 1.20.6. Any updates when 20.6 will be released?

cdobbyn:
@ethanfrogers Any possibility of a 1.20.6 soon™?

kevinawoo:
fyi for anyone looking at this from the future wonder if it got in. [It did in 1.20.6! 🙌 🙌 ](https://spinnaker.io/community/releases/versions/1-20-6-changelog#fixes-1)

PR #855, opened 2020-06-04
refactor(*): Replace Hystrix with Resilience4j

robzienert:


front50-core/src/main/java/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -202,29 +208,23 @@ public long getHealthIntervalMillis() {
...
-          .execute();
-    } catch (HystrixRuntimeException e) {
-      // This handles the case where the hystrix command times out.
-      if (e.getFallbackException() instanceof NotFoundException) {
-        throw (NotFoundException) e.getFallbackException();
-      } else {
-        throw e;
-      }
-    }
+    CircuitBreaker breaker =
```
cfieber, line 211: honestly wonder if a circuit breaker even makes sense here as opposed to just a retry with fallback

if we do want to continue with a circuit breaker, I wonder if we shouldn't have some way of indicating what the backing storage is and adding the breaker around that (e.g sql vs s3 vs ???) rather than one per storage service. (break on which specific ObjectKeyLoader is being used?)

I'm okay with going ahead with this as is since it just preserves the existing behavior but if we want to rethink it a bit happy to discuss it too

robzienert: I'm happy to change what we're doing.

I had originally wired everything up with a retry, but then rolled that back to have one-for-one replacement - I do feel like a retry would be a better fit here. I was also surprised that it wasn't based on backend but just the DAO.

ezimanyi: It also interestingly looks like this logic doesn't distinguish the type of error.  Would we want to retry on a 404?  And likewise would we want enough 404's to trip the circuit breaker?  Again, this looks to be unchanged by this refactor but worth thinking about if we do rethink this.

robzienert: Speaking only for the object store backends, I think we'd want to retry on 404's since their APIs are eventually consistent. I wouldn't want to retry in the case of SQL, though.

ajordens: fwiw the eventual consistency backing stores (namely versioned s3) is a very poor experience ... the breaker is likely unnecessary here _but_ I'd probably just roll with it for the 1:1 replacement.

PR #859, opened 2020-06-08
fix(pipeline): Fix regression in Java conversion of PipelineController

jervi:
Front50 can potensially crash when trying to save a pipeline because `regenerateCronTriggerIds` (if set) is already a boolean, and can't be cast to a String.
Regression introduced in #841.


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -137,8 +151,7 @@ public Pipeline save(@RequestBody Pipeline pipeline) {
     pipeline = ensureCronTriggersHaveIdentifier(pipeline);
 
     if (Strings.isNullOrEmpty(pipeline.getId())
-        || Boolean.parseBoolean(
-            (String) pipeline.getOrDefault("regenerateCronTriggerIds", "false"))) {
+        || (boolean) pipeline.getOrDefault("regenerateCronTriggerIds", false)) {
```
cfieber, line 154: maybe worth an `objectMapper.convertValue(pipeline.getOrDefault("regenerateCronTriggerIds", false), Boolean.class)` unless we are totally sure that these don't come in as a string at some point.. (feels dirty but you see some crazy things in user submitted data)

PR #862, opened 2020-06-10
feat(name validation): Add app name validation

marchello2000:
In case a user submits invalid application name via the API (instead of `deck`), we want to catch bad application names (validation that's in `deck` already)
The validation regex is supplied by config:
```yaml
validation:
  appNameValidator:
    nameRegEx: "^.*$"
```

By default, no validation will be performed (same as today)

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/validator/ApplicationNameValidator.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/validator/ApplicationNameValidator.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


front50-core/src/main/java/com/netflix/spinnaker/front50/validator/AppNameValidator.java
```diff
@@ -0,0 +1,50 @@
...
+    if (Strings.isNullOrEmpty(validationRegEx)) {
+      return;
+    }
+
+    String appName = Optional.ofNullable(application.getName()).orElse("");
+    if (!appName.matches(validationRegEx)) {
+      validationErrors.rejectValue(
+          "name",
+          "application.name.invalid",
+          "Application name can only contain a-z, 0-9, dots, and dashes");
```
cfieber, line 47: should this message reference the validation regex instead ?

marchello2000: doh, thanks! i changed to regex and then didn't update the message

front50-core/src/main/java/com/netflix/spinnaker/front50/validator/AppNameValidator.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AppNameValidator implements ApplicationValidator {
+  private String validationRegEx;
+
+  @Autowired
+  public AppNameValidator(
+      @Value("${validation.app-name-validator.name-reg-ex:#{null}}}") String validationRegEx) {
```
cfieber, line 32: I know its a bit of a pain in the ass, but if this can be in a ConfigurationProperties rather than an Value it sets us up to better document configuration

marchello2000: you're right, i was being lazy... will fix!

PR #866, opened 2020-06-15
chore(dependencies): Autobump korkVersion

spinnakerbot:


ezimanyi:
@robzienert : Mind taking a look at this one?  It looks like it's just a remaining reference to `kork-hystrix` in `front50` (potentially just in the gradle config even, not in actual code).

PR #868, opened 2020-06-22
fix(authz): Activate the RunAsUserToPermissionsMigration again

jervi:
Because the migration is already feature flagged, I think there is no need to expire this migration by using a hard coded date.


cfieber:
@jervi was this no longer necessary or closed because it took too long to more (apologies if the later).

jervi:
I probably just cleaned up my branch without checking if the PR was actually merged. Sorry!

PR #769, opened 2020-04-14
fix(serviceAccounts): Delete managed service account when pipeline is deleted

adriengentil:
Sometime ago, we noticed the sync operation of Fiat was very slow (it was taking between 2 and 3 minutes) and in some cases, the Fiat instance went out of memory. After investigations, we found out that we had about 2000 managed service accounts not associated to any pipelines, deleting them solved our issue with Fiat.

This change deletes the managed service account liked to the pipeline currently being deleted.

I decided to implement the deletion logic of the managed service accounts in Front50 instead of Orca mostly because the orchestration of the deletion of an application is handled by Front50 and it has the knowledge of which pipeline is being removed during the operation.

PR #872, opened 2020-06-24
feat(serviceaccounts): Delete dangling service accounts migration

parkerlarry:
Earlier versions of Front50 did not delete managed service accounts when the corresponding pipeline or application was deleted. This was fixed in PR #769, but that change was not retroactive. This PR adds a migration that will cleanup managed service accounts that are not in use (dangling).

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2020 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.front50.migrations;
+
+import com.netflix.spinnaker.front50.config.annotations.ConditionalOnAnyProviderExceptRedisIsEnabled;
+import com.netflix.spinnaker.front50.model.pipeline.PipelineDAO;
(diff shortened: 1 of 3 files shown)


jervi:
@adriengentil PTAL

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Override
+  public boolean isValid() {
+    return LocalDate.now().isBefore(VALID_UNTIL);
+  }
+
+  @Override
+  public void run() {
+    log.info(
+        "Starting deletion of dangling service accounts ({})", this.getClass().getSimpleName());
```
jervi, line 62: Maybe add a similar info block at the end that notifies of completion?

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
```diff
@@ -0,0 +1,88 @@
...
+                        .distinct())
+            .collect(Collectors.toSet());
+
+    serviceAccountDAO
+        .all()
+        .parallelStream()
+        .map(ServiceAccount::getName)
+        .filter(isManagedServiceAccount())
+        .filter(serviceAccount -> !serviceAccountsToKeep.contains(serviceAccount))
+        .forEach(serviceAccountDAO::delete);
```
jervi, line 82: I'd consider adding a
```
.peek(serviceAccount -> log.info("Deleting managed service account '{}' because it is not in use")
```
just before the `forEach` here.

adriengentil:
I agree with @jervi to have more logs about what is deleted, apart from that LGTM :+1:

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
parkerlarry: fixed 👌

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/DeleteDanglingServiceAccountsMigration.java
parkerlarry: fixed 👌

jervi:
Looks good, however I think you accidentally deleted a few files in the .idea directory at the same time. These are actually supposed to be checked in.

PR #888, opened 2020-07-07
feat(echo): Add dependency on echo

christopherthielen:
This is in prep for moving plugin events from igor to front50

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
@@ -17,4 +17,5 @@
 
 package com.netflix.spinnaker.front50.echo;
 
-public abstract class Event {}
+/** A marker interface for outgoing Events to Echo */
+public interface Event {}
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.front50.echo;
+
+public abstract class Event {}
```
robzienert, line 20: Do we need?

front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Notification.java
```diff
@@ -0,0 +1,47 @@
...
+  @NonNull Map<String, ?> additionalContext;
+
+  @Data
+  public static class Source {
+    @NonNull String application;
+  }
+
+  public enum Type {
+    EMAIL
+  }
```
robzienert, line 41: Wondering aloud: Do we want this to be an enum? If Echo creates a new notification type, it means that we'll need to also update this enum as well.

front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Event.java
christopherthielen: I lifted this from the -nflx code.  I think the point was to have a supertype for `EchoService`'s `String postEvent(@Body Event event);` instead of typing `postEvent` as, for example `PluginEvent` or `Object`.   This could probably be a marker interface instead.

front50-core/src/main/java/com/netflix/spinnaker/front50/echo/Notification.java
christopherthielen: In `echo`, this is actually typed as a `String` (the enum was removed in March by Clay because of plugins).  I don't feel strongly that this should be an enum... it would make as much sense for it to be a constant value.  I'll update.

christopherthielen: Actually, on second thought, I think I do prefer this as an enum.  I think it conveys the types of notifications that Front50 will send, not the set of types that Echo supports.  Changing to a constant felt more awkward, and putting it inline when creating the notification felt more error-prone.

PR #897, opened 2020-07-09
fix(application): Prevent NPE when there are no clusters listed in the project while removing an app from the project

vigneshm:


front50-core/src/main/java/com/netflix/spinnaker/front50/model/application/ApplicationService.java
```diff
@@ -171,12 +171,13 @@ private void removeApplicationFromProjects(String appName) {
...
-                      c ->
-                          Optional.ofNullable(c.getApplications())
-                              .ifPresent(apps -> apps.remove(appName)));
+              Optional.ofNullable(p.getConfig().getClusters())
+                  .ifPresent(
+                      clusters ->
+                          clusters.forEach(
+                              c ->
+                                  Optional.ofNullable(c.getApplications())
+                                      .ifPresent(apps -> apps.remove((appName)))));
```
vigneshm, line 180: No idea if this is the idiomatic way of doing this in Java. Please let me know if there's a better way!

PR #900, opened 2020-07-13
fix(azure storage container): update permission to Private

ishagarg01:
Front50 to switch off Public access on Azure Blob Containers #5892

Issue Raised for the same #5892

### DIFF
--- a/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
+++ b/front50-azure/src/main/java/com/netflix/spinnaker/front50/model/AzureStorageService.java
@@ -64,6 +64,9 @@ public class AzureStorageService implements StorageService {
       } catch (Exception e) {
         // log exception
         blobContainer = null;
+        log.error(
+          "Error encountered while creation blob container for storage in azure",
+          value("exception", e.getMessage()));
       }
     }
     return blobContainer;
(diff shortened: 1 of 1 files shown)


e96wic:
Hey @IshaGarg2504, thank you for the fix. If possible, you could add some log output in the catch block to the error is not silent. I don't really know the codebase very well, but maybe it makes sense to even throw the error again.

zhouchong90:
Please get this merged!! We have a same issue where our organization prohibited the public access on the container. It doesn't need the public access either since it already has the master key.

zhouchong90:
Oh. btw, these 4 lines doesn't need to exist in my opinion. It should leave the setting to user to configure it on the cloud themselves. I'm concerned that sometimes Azure might prevent clients from changing the access level, which will go to the catch block and return null.

BlobContainerPermissions permissions = new BlobContainerPermissions();
        permissions.setPublicAccess(BlobContainerPublicAccessType.CONTAINER);
        permissions.setPublicAccess(BlobContainerPublicAccessType.OFF);
        blobContainer.uploadPermissions(permissions);

robzienert:
Tests are failing. Can you update?

robzienert:
Will merge when things are passing.

pedrohdz:
@e96wic, @robzienert, @zhouchong90,

I implemented a slightly different approach in the following PR:

- https://github.com/spinnaker/front50/pull/1018

The idea is to leave the permissions alone on existing storage containers.  This would allow for backwards compatibility, not locking out front50.  It also allows for users to specify their own permissions when they create the storage container outside of Spinnaker.

The change also defaults to making the storage container private, a sane default IHMO.

jkschneider:
Duplicate of #1018

PR #901, opened 2020-07-15
fix(kork): updates to kork 7.54.0

cfieber:
fixes some breaking changes and dependency updates from latest kork release


front50-s3/front50-s3.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
   implementation "com.netflix.spinnaker.kork:kork-security"
   implementation "com.amazonaws:aws-java-sdk-s3"
   implementation "com.amazonaws:aws-java-sdk-sts"
+  implementation "com.netflix.eureka:eureka-client"
```
cfieber, line 28: this optionally uses an ApplicationInfoManager bean if available (kork-eureka will autoconfigure one of those if enabled)

cfieber:
closes #896

PR #903, opened 2020-07-17
feat(plugins): Add changeSetUrl field to PluginInfo Release

jonsie:


christopherthielen:
@jonsie  worth adding this to the `PluginEvent` and also the trigger payload in echo?

robzienert:
Ping.

PR #915, opened 2020-07-31
feat(pipeline): (optionally) validate that a saved pipeline is not stale

christopherthielen:
i.e., optimistic lock

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
@@ -143,9 +143,12 @@ public class PipelineController {
   @PreAuthorize(
       "@fiatPermissionEvaluator.storeWholePermission() and hasPermission(#pipeline.application, 'APPLICATION', 'WRITE') and @authorizationSupport.hasRunAsUserPermission(#pipeline)")
   @RequestMapping(value = "", method = RequestMethod.POST)
-  public Pipeline save(@RequestBody Pipeline pipeline) {
+  public Pipeline save(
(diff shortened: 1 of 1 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -275,6 +286,20 @@ private PipelineTemplateDAO getTemplateDAO() {
                 "Pipeline Templates are not supported with your current storage backend"));
   }
 
+  private void checkForStalePipeline(Pipeline pipeline, GenericValidationErrors errors) {
+    Pipeline existingPipeline = pipelineDAO.findById(pipeline.getId());
```
cfieber, line 290: I feel like there is a case here where there isn't an existing pipeline (via the save entrypoint), which should probably be legit from a stale checking perspective. I also suspect the submitted payload in that case would not have a last modified (unless that is magically set somewhere along the way with a default value).

christopherthielen: My intent was to use this for the deck save pipeline dialog, in which case it would only request stale checking for existing pipelines.  But yeah, might as well make it more robust.  I'm thinking if the submitted payload has an ID (i.e., it's an update op, not an insert) and stale check is requested, we can bomb out in surprising scenarios like the existing pipeline not found.

> unless that is magically set

I think `lastModified` is set when the entity is actually persisted, so yeah good call.

christopherthielen:
@cfieber thanks for looking, I pushed a new commit https://github.com/spinnaker/front50/pull/915/commits/23039ff74cd979d973d7f85658524870c9eab75e

PR #916, opened 2020-08-03
fix(misc): Recover application permission also when we undo delete op…

srekapalli:
- Currently for recovering a deleted application we need to call `/recover` endpoint for application and application permission separately and if missed to call for permission, we won't be able to perform any write operation on the application.  With this change, we recover permission record also along with application record.

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/AdminController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/AdminController.java
@@ -43,8 +43,7 @@ public class AdminController {
   @RequestMapping(value = "/recover", method = RequestMethod.POST)
   void recover(@RequestBody AdminOperations.Recover operation) {
     adminOperations.forEach(o -> o.recover(operation));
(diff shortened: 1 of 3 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/AdminController.java
```diff
@@ -42,5 +43,15 @@ public AdminController(Collection<AdminOperations> adminOperations) {
   @RequestMapping(value = "/recover", method = RequestMethod.POST)
   void recover(@RequestBody AdminOperations.Recover operation) {
     adminOperations.forEach(o -> o.recover(operation));
+    // If we are recovering application , recover application permission also. Without that
```
ajordens, line 46: Suggested wording change (Also avoids the ` , ` which looks a little strange in the comment).

```
// Application permissions need to be recovered alongside the application itself
```

In the future, it can be a little easier to read if we line break on full sentences.

```
// If we are recovering application, recover application permission also.
// Without that subsequent CRUD operations on application will fail.
```

PR #917, opened 2020-08-04
feat(plugins): Add remote extensions field to a plugin info release

jonsie:
In https://github.com/spinnaker/orca/pull/3840 I realized I had the cart before the horse on a few items, specifically the basic remote extension point configuration.  This will not only allow me to get away from abusing the PreconfiguredJob stuff even more (by allowing for specific remote stage configuration), but also lay some of the ground work for other types of remote extension points.

Adds a `remoteExtensions` field to a plugin info `release`.  A plugin that is just a remote extension look like this:

```
{
  "id": "netflix.remote",
  "description": "A remote plugin",
  "provider": "csmalley",
  "releases": [
    {
      "version": "0.0.1",
      "date": "2020-06-16T02:20:14.889Z",
      "requires": "orca>=1.0.0",
      "url": null,
      "sha512sum": null,
      "preferred": false,
      "lastModified": "2020-06-18T23:08:48.899Z",
      "lastModifiedBy": "csmalley",
      "remoteExtensions": [
        {
          "type": "stage",
          "id": "netflix.remote.remoteWait",
          "transport": {
            "http": {
              "url": "http://service.com/remoteWait",
              "config": {}
            }
          },
          "config": {
            "type": "remoteWait",
            "label": "Wait on a remote service",
            "description": "A stage that waits on a remote service",
            "parameters": {
              "waitTime": 30,
              "message": "done!"
            }
          }
        }
      ]
    }
  ]
}
```

The idea here is that we use the `requires` field in-conjunction with the remote extension `type` to determine which service to configure with the specific remote extension and corresponding remote extension configuration.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
@@ -10,12 +10,17 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import org.hibernate.validator.constraints.URL;
 
 /**
  * A Spinnaker plugin's remote extension configuration.
  *
  * <p>This model is used by Spinnaker to determine which extension points and services require
  * remote extension point configuration.
+ *
+ * <p>The plugin release {@link PluginInfo.Release#requires} field is used to inform Spinnaker which
+ * service to use in configuring the extension point {@link #type} and additionally if the remote
+ * extension is compatible with the running version of the Spinnaker service.
  */
 @Data
 @NoArgsConstructor
@@ -23,13 +28,11 @@ public class RemoteExtension {
 
   /**
    * The remote extension type. The remote extension is configured in the service that implements
-   * this extension type. The plugin release {@link PluginInfo.Release#requires} field is used to
-   * inform Spinnaker which service to use in configuring the remote extension type and if the
-   * remote extension is compatible with the running version of the Spinnaker service.
+   * this extension type.
    */
   @Nonnull private String type;
 
-  /** Identifier of the remote extension. Used for observability. */
+  /** Identifier of the remote extension. Used for tracing. */
   @Nonnull private String id;
 
   /**
@@ -61,7 +64,7 @@ public class RemoteExtension {
     public static class Http {
 
       /** URL for remote extension invocation. */
-      @Nonnull private String url;
+      @URL @Nonnull private String url;
 
       /** A placeholder for misc. configuration for the underlying HTTP client. */
       @Nonnull private Map<String, Object> config = new HashMap<>();
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
@@ -1,0 +1,70 @@
+package com.netflix.spinnaker.front50.model.plugins.remote;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.netflix.spinnaker.front50.model.plugins.PluginInfo;
+import com.netflix.spinnaker.front50.model.plugins.remote.stage.StageRemoteExtensionConfig;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A Spinnaker plugin's remote extension configuration.
+ *
+ * <p>This model is used by Spinnaker to determine which extension points and services require
+ * remote extension point configuration.
+ */
+@Data
+@NoArgsConstructor
+public class RemoteExtension {
+
+  /**
+   * The remote extension type. The remote extension is configured in the service that implements
+   * this extension type. The plugin release {@link PluginInfo.Release#requires} field is used to
+   * inform Spinnaker which service to use in configuring the remote extension type and if the
+   * remote extension is compatible with the running version of the Spinnaker service.
+   */
+  @Nonnull private String type;
+
+  /** Identifier of the remote extension. Used for observability. */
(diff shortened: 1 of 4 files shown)


front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,70 @@
...
+   * with and the necessary configuration.
+   */
+  @Nonnull private RemoteExtensionTransport transport = new RemoteExtensionTransport();
+
+  /** Configures the remote extension point. */
+  @JsonTypeInfo(
+      use = JsonTypeInfo.Id.NAME,
+      include = JsonTypeInfo.As.EXTERNAL_PROPERTY,
+      property = "type")
+  @JsonSubTypes({@JsonSubTypes.Type(value = StageRemoteExtensionConfig.class, name = "stage")})
```
jonsie, line 46: `stage` here corresponds to the `RemoteExtension#type` field.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Data
+    @NoArgsConstructor
+    public static class Http {
+
+      /** URL for remote extension invocation. */
+      @Nonnull private String url;
+
+      /** A placeholder for misc. configuration for the underlying HTTP client. */
+      @Nonnull private Map<String, Object> config = new HashMap<>();
```
jonsie, line 67: This can be flushed out over time, but for now it's a catch-all for any underlying client configuration we need.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/stage/StageRemoteExtensionConfig.java
```diff
@@ -0,0 +1,25 @@
...
+  @Nonnull private String type;
+
+  /** Label to use on the Spinnaker UI while configuring pipeline stages. */
+  @Nonnull private String label;
+
+  /** Description to use on the Spinnaker UI while configuring pipeline stages. */
+  @Nonnull private String description;
+
+  /** Map of stage parameter names and default values. */
+  @Nonnull private Map<String, Object> parameters = new HashMap<>();
```
jonsie, line 24: I think for now a map of the parameter name and the default value will suffice.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * The remote extension type. The remote extension is configured in the service that implements
+   * this extension type. The plugin release {@link PluginInfo.Release#requires} field is used to
+   * inform Spinnaker which service to use in configuring the remote extension type and if the
+   * remote extension is compatible with the running version of the Spinnaker service.
+   */
+  @Nonnull private String type;
+
+  /** Identifier of the remote extension. Used for observability. */
```
robzienert, line 32: nit: s/observability/tracing/

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+@Data
+@NoArgsConstructor
+public class RemoteExtension {
+
+  /**
+   * The remote extension type. The remote extension is configured in the service that implements
+   * this extension type. The plugin release {@link PluginInfo.Release#requires} field is used to
+   * inform Spinnaker which service to use in configuring the remote extension type and if the
+   * remote extension is compatible with the running version of the Spinnaker service.
```
robzienert, line 28: Great docs, but I think some of this actually belongs at the top-level class docs, specifically where you start talking about `requires` and you're no longer talking about the `type` property, but how we configure the said `type`. Hope that makes sense?

front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,70 @@
...
+  @JsonTypeInfo(
+      use = JsonTypeInfo.Id.NAME,
+      include = JsonTypeInfo.As.EXTERNAL_PROPERTY,
+      property = "type")
+  @JsonSubTypes({@JsonSubTypes.Type(value = StageRemoteExtensionConfig.class, name = "stage")})
+  @Nullable
+  public RemoteExtensionConfig config;
+
+  /** Root remote extension configuration type. */
+  public interface RemoteExtensionConfig {}
```
robzienert, line 51: Rather than define the extension point config type in front50 (making front50 changes required for most remote extension), I think we should investigate a shared model or service-exposed schema for extension point configs - personally I favor the schema idea:

For any extension point that has a config, say `MyRemoteExtensionConfig`, we could convert to JSON Schema (or similar) and expose it from something like `/plugins/extension-point-configs/`. Each config would have a `type`, as you have it, but structured like `{service}/{configType}[@version?]`[1], allowing front50 to lookup the correct schema when a remote plugin is being published and we want to do some type / constraint validation. This would let us continue to define, in a single place, everything for an extension point.

I don't think we need to change the PR as it exists today, since this is talking about a lot of work - we can always retrofit what you have here after we've got something working end-to-end. 💯 

[1] Not sure if version is needed, but probably? 🤷 

One problem with this is that services tend to have dependencies _on_ front50, but not the other way around. I think I'd also be OK with services publishing their schemas into front50 on startup. 🤷

jonsie: Super interesting idea. 

I think a version would be needed.  Each change to the config schema would need to be versioned, and then that version would need to be the published version into the front50 plugin info release.  This is because we will be rolling out remote extensions by version, and we can't really swap out the underlying config schema without at least rolling out a new version.

christopherthielen: What would schema validation failures gate? Just publishes of new plugins?

PR #920, opened 2020-08-06
feat(plugins): Delete RemoteExtension and StageRemoteExtensionConfig and use model from kork-api

jonsie:
Just getting our feet wet here with a shared model pattern - about as basic as it gets.


front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/remote/stage/StageRemoteExtensionConfig.java
```diff
@@ -1,25 +0,0 @@
...
-
-  /** Label to use on the Spinnaker UI while configuring pipeline stages. */
-  @Nonnull private String label;
-
-  /** Description to use on the Spinnaker UI while configuring pipeline stages. */
-  @Nonnull private String description;
-
-  /** Map of stage parameter names and default values. */
-  @Nonnull private Map<String, Object> parameters = new HashMap<>();
-}
```
jonsie, line 25: This remote extension config will be moved to Orca.

PR #930, opened 2020-08-19
feat(gitRepoArtifact): include migration to move subpath to location field

edgarulg:
For `git/repo` artifacts, we include the subPath value as metadata, this cause some issues when a user configure a pipeline with expected artifacts with type git/repo, same URL and version but different subPath value and fails when tries to match artifact to use.

This PR is about create a Front50 migration to get the saved pipelines and only for expected artifacts with type `git/repo` and a subpath defined, move the subpath value to the location field.

Currently, there are PR to handle backwards compatibility in Clouddriver and Deck:
https://github.com/spinnaker/clouddriver/pull/4815
https://github.com/spinnaker/deck/pull/8507


front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+
+    when:
+    migration.run()
+
+    then:
+    _ * pipelineDAO.all() >> { return [pipeline] }
+    0 * pipelineDAO.update("pipeline-1", _)
+
+    where:
+    original  || _
```
danielpeach, line 45: This syntax is a little unusual to me (are you taking this approach because you need at least two columns in the table?) If you modeled it after other tests in this project are structured, that's fine.

Otherwise, you can structure it like this, I think:
```
original << [
   [:],
   [type: "bitbucket/file", reference: "test-ref"],
   ...
]
```

front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+  def "should set subpath in location"() {
+    given:
+    def pipeline = new Pipeline([application: "test", expectedArtifacts: [[matchArtifact: original]]])
+    pipeline.id = "pipeline-1"
+
+    when:
+    migration.run()
+
+    then:
+    pipeline.get("expectedArtifacts").find {true}.matchArtifact.location == "test-subpath"
```
danielpeach, line 66: is this the same as `pipeline.get("expectedArtifacts").first().matchArtifact.location == ...`?

front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+    where:
+    original  || _
+    [type: "git/repo", reference: "test-ref", version: "test-ver", metadata: [subPath: "test-subpath"]] || _
+    [type: "git/repo", reference: "test-ref", version: "test-ver", location: "", metadata: [subPath: "test-subpath"]] || _
+  }
+
+  @Unroll
+  def "should set subpath to location with matchArtifact and defaultArtifact"() {
+    given:
+    def pipeline = new Pipeline([application: "test", expectedArtifacts: [[matchArtifact: match, defaultArtifact: defaul]]])
```
danielpeach, line 79: s/defaul/default

front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+    migration.run()
+    migration.run() // subsequent migration run should be a no-op
+
+    then:
+    2 * pipelineDAO.all() >> { return [pipeline] }
+    1 * pipelineDAO.update("pipeline-1", _)
+    0 * _
+
+    where:
+    match  || defaul
```
danielpeach, line 92: Can you please align the columns?

front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
edgarulg: Yes, it is, I just wanted to check the first element. I changed it to use the `first` method instead.

front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy
edgarulg: I changed it for a more convenient name because `default` is a reserved word.

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigration.java
```diff
@@ -0,0 +1,132 @@
...
+            return false;
+          }
+
+          return expectedArtifacts.stream()
+              .anyMatch(
+                  e -> {
+                    Map matchArtifact = (Map<String, Object>) e.get("matchArtifact");
+                    Map defaultArtifact = (Map<String, Object>) e.get("defaultArtifact");
+
+                    return matchArtifact != null
```
ezimanyi, line 70: It would make this easier to read if there were a function:
```boolean shouldMigrateArtifact(@Nullable Map<String, Object> artifact)```
that encapsulated these three checks as they appear in sequence 4 times in this migration.

File: front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigration.java [hunk 6]
```diff
@@ -0,0 +1,132 @@
...
+    log.info(
+        "Move git/repo subpath to location field (application: {}, pipelineId: {}, expectedArtifacts: {})",
+        pipeline.getApplication(),
+        pipeline.getId(),
+        pipeline.get("expectedArtifacts"));
+
+    List<Map<String, Object>> expectedArtifacts =
+        (List<Map<String, Object>>) pipeline.get("expectedArtifacts");
+    for (Map<String, Object> expectedArtifact : expectedArtifacts) {
+      Map<String, Object> matchArtifact =
```
ezimanyi, line 110: Similar to the above comment, this would be more readable if there were a function
```void migrateArtifact(@Nullable Map<String, Object> artifact)```
so we don't have the logic repeated for match/expected.

File: front50-migrations/src/test/groovy/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigrationSpec.groovy [hunk 7]
```diff
@@ -0,0 +1,100 @@
...
+    where:
+    original << [
+      [:],
+      [type: "bitbucket/file", reference: "test-ref"],
+      [type: "github/file", reference: "test-ref", version: "test-ver"],
+      [type: "git/repo", reference: "test-ref", version: "test-ver"],
+      [type: "git/repo", reference: "test-ref", version: "test-ver", location: "test-subpath"],
+      [type: "git/repo", reference: "test-ref", version: "test-ver", metadata: [foo: "test-foo"]],
+      [type: "git/repo", reference: "test-ref", version: "test-ver", location: "test-subpath", metadata: [foo: "test-foo"]],
+      [type: "git/repo", reference: "test-ref", version: "test-ver", location: "test-subpath", metadata: [subPath: "test-subpath"]]
```
ezimanyi, line 53: Good test cases 👍

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MoveSubpathToLocationMigration.java
```diff
@@ -0,0 +1,132 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MoveSubpathToLocationMigration implements Migration {
+
+  private static final Logger log = LoggerFactory.getLogger(CloudProvidersStringMigration.class);
+
+  // Only valid until October 1st, 2020
+  private static final Date VALID_UNTIL = new GregorianCalendar(2020, 10, 1).getTime();
```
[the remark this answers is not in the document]
ezimanyi: Also, I'd strongly recommend passing in `Calendar.OCTOBER` for the month (or whichever month you choose)...as this is actually November 1st, 2020 because the month argument is indexed from 0.

PR #934, opened 2020-08-25
fix(gcs): Fix startup error when bucketLocation is not specified

ezimanyi:
Fixes spinnaker/spinnaker#6000.

* test(gcs): Add some implementations to FakeStorageRpc

  In an upcoming commit, we'll need to also support getting a bucket; to prepare for that let's update the FakeStorageRpc to:
  * Keep track of what bucket objects are in (rather than now where it assumes they are all in the same bucket)
  * Handle creating and getting a bucket

  We'll add an additional Map around the blobs data structure that maps the bucket to a Map of blobs.  (This is a bit of a complex data structure that in production code might be better expressed as classes, but for this single fake in a test it's porbably fine.)

  Creating a bucket initializes its blobs to an empty map; we know if a bucket exists by whether it has an entry in the map.

  I used the documentation on the interface we're mocking to decide how to handle a missing bucket; in some cases we just return null
  (when getting an object or a bucket) while in others we propagate the 404.

  This required one small change to the existing tests, which is that we'll need to create the bucket before we use them in the tests.

* test(gcs): Add test to demonstrate NPE on missing bucketLocation

  This commit adds a broken test to demostrate the NPE that occurs on startup when the bucketLocation is not specified.

* fix(gcs): Fix startup error when bucketLocation is not specified

  There was a regression in 1.22 where omitting bucketLocation from one's GCS config now causes an error on startup. Prior to the rewrite of GcsStorageService, we accepted either null or empty string as the bucket location; now that GcsStorageService is in kotlin and does not use a String? for the field, an error occurs on trying to create the GcsStorageService.

  The fix is just to default the string to "" in the config properties.

### DIFF
--- a/front50-gcs/src/test/kotlin/com/netflix/spinnaker/front50/model/FakeStorageRpc.kt
+++ b/front50-gcs/src/test/kotlin/com/netflix/spinnaker/front50/model/FakeStorageRpc.kt
@@ -57,11 +57,12 @@ internal class FakeStorageRpc(private val clock: Clock) : StorageRpc {
 
   private data class Blob(val storageObject: StorageObject, val content: ByteArray)
 
-  private val blobs = mutableMapOf<String, MutableList<Blob>>()
+  private val buckets = mutableMapOf<String, MutableMap<String, MutableList<Blob>>>()
 
   override fun create(storageObject: StorageObject, data: InputStream, options: MutableMap<Option, *>): StorageObject {
     if (options.isNotEmpty()) throw UnsupportedOperationException("unsupported options to create: ${options.keys}")
     if (storageObject.generation != null) throw UnsupportedOperationException("can't call create with a specific generation")
+    val blobs = buckets[storageObject.bucket] ?: throw StorageException(404, "bucket ${storageObject.bucket} does not exist")
     val generations = blobs.getGenerations(storageObject)
     val stampedObject = storageObject.clone()
     stampedObject.generation = generations.size + 1L
@@ -73,6 +74,7 @@ internal class FakeStorageRpc(private val clock: Clock) : StorageRpc {
   override fun patch(storageObject: StorageObject, options: MutableMap<Option, *>): StorageObject {
     if (options.isNotEmpty()) throw UnsupportedOperationException("unsupported options to patch: ${options.keys}")
     if (storageObject.generation != null) throw UnsupportedOperationException("can't call patch with a specific generation")
+    val blobs = buckets[storageObject.bucket] ?: throw StorageException(404, "bucket ${storageObject.bucket} does not exist")
     val generations = blobs.getGenerations(storageObject)
     if (generations.isEmpty()) throw StorageException(404, "no object ${fullPath(storageObject)}")
     val foundObject = generations[generations.size - 1].storageObject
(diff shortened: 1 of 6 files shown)


ezimanyi:
Added the `backport-candidate` label as this was a regression in 1.22 caused by #911.  I've written tests demonstrating the bug and the fix, so this should satisfy the [release branch patch criteria](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria).

front50-gcs/src/test/kotlin/com/netflix/spinnaker/front50/model/FakeStorageRpc.kt
```diff
@@ -57,11 +57,12 @@ internal class FakeStorageRpc(private val clock: Clock) : StorageRpc {
 
   private data class Blob(val storageObject: StorageObject, val content: ByteArray)
 
-  private val blobs = mutableMapOf<String, MutableList<Blob>>()
+  private val buckets = mutableMapOf<String, MutableMap<String, MutableList<Blob>>>()
```
maggieneterval, line 60: > We'll add an additional Map around the blobs data structure that maps the bucket to a Map of blobs. 

This is maybe a just-me-problem because I've never looked at any of this code until this morning, but if we aren't going to type this with a class can we either include this description from your commit message as a comment here, or rename `buckets` --> `bucketsToBlobs` or something that makes it a bit clearer what's in this nested data structure?

ezimanyi: Good call, looking at this again today it's less clear than it was to yesterday Eric.  Let me take a look at whether it's clearer to add a comment or just make a tiny inner class for this.

ezimanyi: Ok, I make a couple of small classes---let me know if this is clearer!

ezimanyi:
* test(gcs): Replace nested map with classes

  To simplify the nested map, define a Buckets class and a BucketContents class, with operation names that make it more clear what is happening.

  Also, I realized that before my changes we were appending the bucket name to the object name before putting it into the map, but now we don't need to do that anymore because we have a separate map for each bucket, so remove some unecessary calls to fullPath (which is now only used in printing error messages).

front50-gcs/src/test/kotlin/com/netflix/spinnaker/front50/model/FakeStorageRpc.kt
maggieneterval: Oh wow this is 💯 ! (But also, sorry, didn't mean for that comment to lead to this much more work for you!)

ezimanyi:
Found one tiny thing I thought I changed while looking at this before merging.  (I made a private constructor that takes a map, then a public constructor that always passes an empty map...but the way to do this in kotlin is just make a `val` that's not part of the constructor at all.). Fixed that and added `ready-to-merge` so I can't change this any more 😀

front50-gcs/src/test/kotlin/com/netflix/spinnaker/front50/model/FakeStorageRpc.kt
ezimanyi: No worries at all, this is way better and I learned some more kotlin along the way!

ezimanyi:
@Mergifyio refresh

german-muzquiz:
@Mergifyio backport release-1.22.x

PR #948, opened 2020-09-15
fix(misc): Fix for application upsert operation when permissions were…

srekapalli:
Here is the scenario where upsertApplication tasks are failing:
-- API user creates a new application along with permissions associated with it. (Orca calls 2 apis , one to create/update app and one to create/update perm)
-- API user deletes the application and the rest of the related entities( permissions, pipelines etc)
-- API user attempts to update the deleted application 
-- Orca  retrieves the application from front50 (Get API)
-- Front50 retrieves the application record from cache (fallback  if not found in DB) and proceeds to update the application
-- During application update process, event listener kicks in and fails to retrieve the permission record  both from DB and cache, because cache refresh cycle kicked in between get and update operation.
-- If the original `Get` call fails to retrieve both from DB and cache, Orca issues a Create API call which succeeds just fine.
-- This fix is to catch those situations where application cache and application permission cache refresh cycles are not in sync causing one to succeed and other to throw not found exception.
-- upsertApplication task calls application permission update API in the process so we will be covered if application update misses to update ChaosMonkey perms.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/listeners/ChaosMonkeyApplicationEventListener.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/listeners/ChaosMonkeyApplicationEventListener.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.front50.ApplicationPermissionsService;
 import com.netflix.spinnaker.front50.config.ChaosMonkeyEventListenerConfigurationProperties;
 import com.netflix.spinnaker.front50.events.ApplicationEventListener;
 import com.netflix.spinnaker.front50.model.application.Application;
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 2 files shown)


ajordens:
How does this permission handling in `front50` around chaos monkey relate to what `ChaosMonkeyApplicationResourcePermissionSource` is doing in `fiat`?

(cc: @jonsie)

marchello2000 (approved):
lgtm

jonsie:
We need to migrate away from using this.  Chaos monkey permissions are no longer set this way (they are now set in Fiat's `ChaosMonkeyApplicationResourcePermissionSource`) - however we can't simply delete this code and call it a day as we could get into a situation where users are locked out of their applications (one of the major design flaws with adding this permission in front50 in the first place, hence why I wrote the `ChaosMonkeyApplicationResourcePermissionSource`).  

All this code does at this moment is write unnecessary data to an application's permissions.

I'll brainstorm on how to migrate away from this.  This bug is *exactly* why doing it this way is bad.

jonsie (approved):
Merge it and deploy it to resolve the immediate issue -- we can then write a front50 `Migration` and determine what to do with Deck too.

PR #949, opened 2020-09-15
feat(test): add front50 test fixture

danielpeach:
This is modeled after https://github.com/spinnaker/orca/tree/master/orca-api-tck and has the same purpose (letting plugin developers write Front50 integration tests).

I named the project `front50-api-tck`, following the pattern in Orca (and I'm opening PRs for test fixtures in the other projects). There's no corresponding `front50-api` subproject, which makes the name seem a little strange.

### DIFF
--- a/front50-api-tck/README.md
+++ b/front50-api-tck/README.md
@@ -1,0 +1,3 @@
+# front50-api-tck
+
+Test harnesses and utilities for Front50 plugin developers.
--- a/front50-api-tck/front50-api-tck.gradle
+++ b/front50-api-tck/front50-api-tck.gradle
@@ -1,0 +1,29 @@
+apply from: "${project.rootDir}/gradle/kotlin.gradle"
+
+dependencies {
+  implementation(project(":front50-web"))
+  implementation("com.h2database:h2")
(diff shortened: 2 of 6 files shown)


jonsie:
Do we want to make front50 pluggable?  I was kinda thinking this might be one of the services that we never enable plugins on.  

Regardless, I'm wondering if we want to merge this before adding `front50-api` - what's the point?  If we then never add `front50-api` we end up with yet another module that is essentially cruft.

danielpeach:
We have plugins that extend [PipelineValidator](https://github.com/spinnaker/front50/blob/master/front50-core/src/main/java/com/netflix/spinnaker/front50/validator/PipelineValidator.java#L23-L29) and we'd prefer to have the test fixture be in a common place in open source.

My memory was that `PipelineValidator` was a true extension point, but of course now I'm seeing that it's not! Dang. In any case, we're using it.

robzienert (approved):

robzienert:
Front50 will have to be pluggable to some extent. We don't want to add / support a bunch of storage engines, for example.

PR #950, opened 2020-09-15
feat(plugins): make Front50 pluggable

danielpeach:
Based on @jonsie's comment in https://github.com/spinnaker/front50/pull/949, I thought "of course Front50 is pluggable, we've written Front50 plugins!". 

It's true that Armory has written Front50 plugins and that they work for some range of Spinnaker versions, but it's also true that Front50 doesn't import `PluginsAutoConfiguration`, which is pretty confusing. 

It turns out that since `kork-plugins` ends up on the [classpath](https://github.com/spinnaker/front50/blob/master/front50-core/front50-core.gradle#L31) and since Front50 `Main` does a component scan of `com.netflix.spinnaker.config`, we used to get `PluginsAutoConfiguration` for free. I think we lost that behavior when the `@Configuration` annotation on `PluginsAutoConfiguration` [got dropped here](https://github.com/spinnaker/kork/pull/752/files#diff-abed1df5ae8a77ddedc0f40f1c6399acL77). Whoops.


robzienert:
I'm not sure dropping that configuration annotation was deliberate. In any case, if this works, then cool.

robzienert (approved):

PR #952, opened 2020-09-17
fix(misc): Conditionally enable migration runner

srekapalli:
- Conditionally enabling migration runner

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
@@ -29,7 +29,9 @@ import org.springframework.stereotype.Component;
 /**
  * Migration runner runs all the registered migrations as scheduled. By default migration runner
  * will <strong>not</strong> run and need to set <strong>`migrations.enabled`</strong> property to
- * enable it. Note: Ideally migrations should be running only on one instance.
+ * enable it.
+ *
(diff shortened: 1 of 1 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
```diff
@@ -22,10 +22,19 @@
...
 
+/**
+ * Migration runner runs all the registered migrations as scheduled. By default migration runner
+ * will <strong>not</strong> run and need to set <strong>`migrations.enabled`</strong> property to
+ * enable it.
+ *
+ * <p>Note: Ideally migrations should be running only on one instance.
+ */
 @Component
+@ConditionalOnProperty(name = "migrations.enabled", havingValue = "true", matchIfMissing = false)
```
marchello2000, line 37: this is reverse behavior from what is there today (default not running). I think this makes sense, and I doubt someone has additional migrations on top of what's already in front50

marchello2000 (approved):
lgtm, thank you!!!

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
srekapalli: There are very few migrations that are active now and some of them have conditional on them directly so on the OSS side we will turn off those. May be better leave it `default running` and turn off explicitly ?

jonsie (approved):

PR #953, opened 2020-09-17
migration(chaosmonkey): Remove chaos monkey user migration

jonsie:
Puts us on a path towards deleting the chaos monkey even listener code too.  Tweaked the config a bit so we can roll this migration out in a simple way.

### DIFF
--- a/front50-migrations/front50-migrations.gradle
+++ b/front50-migrations/front50-migrations.gradle
@@ -17,6 +17,7 @@
 dependencies {
   implementation project(":front50-core")
 
+  implementation "com.netflix.spinnaker.fiat:fiat-core:$fiatVersion"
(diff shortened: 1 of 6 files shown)


front50-migrations/front50-migrations.gradle
```diff
@@ -17,6 +17,7 @@
 dependencies {
   implementation project(":front50-core")
```
srekapalli, line 19: You need this for ?

jonsie: `com.netflix.spinnaker.fiat.model.resources.Permissions` comes from fiat-core.

srekapalli (approved):
nice. 👍

PR #961, opened 2020-09-25
chore(dependencies): Autobump fiatVersion

spinnakerbot:


jonsie:
Bumping over here: https://github.com/spinnaker/front50/pull/962

PR #986, opened 2020-10-29
fix(bug): Reject blank strings for pipeline name, application and als…

srekapalli:
- During create operations we are allowing `blank` strings for required attrs like (id, name and application) and the fix here is to reject null, empty and blank values for any of the required attrs.


jonsie (approved):
LGTM

ajordens (approved):

PR #987, opened 2020-10-30
fix(web): Migrate controller tests from S3 to SQL

ajordens:
Of course the S3-backed tests never actually ran
and the migration to SQL exposed a handful of
minor bugs/inconsistencies.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
@@ -220,14 +220,15 @@ public abstract class StorageServiceSupport<T extends Timestamped> {
         SupplierUtils.recover(
             () -> service.loadObject(objectType, buildObjectKey(id)),
             e ->
-                allItemsCache.get().stream()
+                Optional.ofNullable(allItemsCache.get()).orElseGet(HashSet::new).stream()
                     .filter(item -> item.getId().equalsIgnoreCase(id))
                     .findFirst()
                     .orElseThrow(
                         () ->
                             new NotFoundException(
                                 String.format(
-                                    "No item found in cache with id of %s", id.toLowerCase()))));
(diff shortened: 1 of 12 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -157,9 +158,11 @@ public Pipeline save(
     if (Strings.isNullOrEmpty(pipeline.getId())
         || (boolean) pipeline.getOrDefault("regenerateCronTriggerIds", false)) {
       // ensure that cron triggers are assigned a unique identifier for new pipelines
-      pipeline.getTriggers().stream()
-          .filter(it -> "cron".equals(it.getType()))
-          .forEach(it -> it.put("id", UUID.randomUUID().toString()));
+      Collection<Trigger> triggers = pipeline.getTriggers();
```
ajordens, line 161: We actually need to do this because `pipeline.getTriggers()` returns an ObjectMapper'd copy.

ajordens:
Some groovy reformatting apparently made it into this PR.

Let me know if it's grief worthy and I can artisanally back it out.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/StorageServiceSupport.java
```diff
@@ -220,14 +220,15 @@ public T findById(String id) throws NotFoundException {
         SupplierUtils.recover(
             () -> service.loadObject(objectType, buildObjectKey(id)),
             e ->
-                allItemsCache.get().stream()
+              Optional.ofNullable(allItemsCache.get()).orElse(new HashSet<>()).stream()
```
robzienert, line 223:
```suggestion
              Optional.ofNullable(allItemsCache.get()).orElseGet(HashSet::new).stream()
```

Small optimization.

robzienert (approved):

PR #988, opened 2020-11-01
feat(pipeline executions/front50) : Added code to save multiple pipelines at once to sql database.

sanopsmx:
feat(pipeline executions/front50) : Added code to save multiple pipelines at once to sql database.

This is part of: spinnaker/spinnaker#6147.

Enhanced PipelineController.java to

Added new rest api method bulksave(List<Map> pipelineList, Boolean staleCheck)
This method accepts a list of pipelines json.
This method checks each and every pipeline for authorization.
If the pipeline has valid authorization, then it will be saved to a savedPipelineList.
If the pipeline do not have a valid authorization, then it will be saved to a errorPipelineList
 and an error message is also populated.
All the savedPipelinesList will be saved to sql database using bulk insert.
This method returns a Map object having the below data:
{
  “Successful”: <count>,
  “Failed”: <cound>,
  “Failed_list”: [<array of failed pipelines - (application, pipeline name, etc) and the error message]
}

Enhanced AuthorizationSupport.java to

Added code to accept the pipeline list and authorize the pipelines.

Enhanced PipelineControllerSpec.java to

Added bean of FiatPermissionEvalutor.java as it is used in the PipelineController constructor.

Enhanced PipelineControllerTck.groovy to

Added bean of FiatPermissionEvalutor.java as it is used in the PipelineController constructor.

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/AuthorizationSupport.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/AuthorizationSupport.java
@@ -87,4 +87,11 @@ public class AuthorizationSupport {
 
     return permissionEvaluator.hasPermission(auth, application, "APPLICATION", "EXECUTE");
   }
+
+  public boolean hasRunAsUserPermission(final List<Pipeline> pipelineList) {
+    for (Pipeline pipeline : pipelineList) {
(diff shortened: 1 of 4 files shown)


sanopsmx:
@robzienert @ajordens ....Can you please review this PR

PR #993, opened 2020-11-11
feat(pipeline_template): Add support of semver for template tags

### LINKED ISSUE #5569: Pipeline Template tags should support SemVer
### Issue Summary:

When saving a Pipeline Template with tags, only certain tags are valid. We would like to tag our Pipeline Templates with SemVer, to allow developers to pin pipeline version.

### Feature Area:

Managed Pipeline Templates v2
Spin CLI v1.14.0

### Description:

Pipeline Template may be saved with a tag. I would expect that functionality to allow for any string, or at least any alphanumeric string. However, it only allows tagging with a small list of approved tags, and there is little to no documentation on what those tags are.

I would like to be able to tag with any string, especially SemVer. This will allow us to push updates to these pipelines, while allowing consuming teams to update at their own pace. 

Additionally, right now there is no error message produced in the Spin CLI if you try to save a template with an invalid tag - it just doesn't save the pipeline. Nor is there any indication in the Spin CLI help that `--tags` only takes a certain set of values, much less what those values might be.

### Steps to Reproduce:

Attempt to save a valid pipeline template using the tags flag with SemVer:

```
spin pipeline-template save --file pipeline.json --tag v0.1.0
```

It will say that it saved successfully, but no template will be saved.

### Additional Details:

The code in question seems to be [here](https://github.com/spinnaker/front50/blob/master/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java#L61).

micnncim:
Resolves https://github.com/spinnaker/spinnaker/issues/5569

This PR would add support of semver for pipeline template tags. Seems many users want such a option in the above issue.

The regex pattern is the almost same as: https://github.com/spinnaker/front50/blob/28206e1d8fae2cc071ad168d9b43e7551de0a808/front50-core/src/main/java/com/netflix/spinnaker/front50/model/plugins/PluginInfo.java#L95 and also supports `v` prefix as an option.

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
@@ -41,6 +41,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.codec.binary.Hex;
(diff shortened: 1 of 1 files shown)


micnncim:
I'll fix the errors later but want feedback on this change. Seems many users want this feature based on ]the referenced issue](https://github.com/spinnaker/spinnaker/issues/5569) above.

micnncim:
@ajordens @jonsie @plumpy 
I’m not sure you’re some of the owners of this file but could you take a look at this PR?

micnncim:
ping

PR #1014, opened 2021-02-16
fix(sql): Ensure that `Application.createTs` is reliably returned when sql is used

ajordens:
For historical reasons, the original `createTs` value on the Application object itself
is not reliably stored.

When sql backends are used, the `created_at` column is the most accurate representation
of the time at which an Application was created.

This PR will override any `createTs` value that happens to exist in the Application
json blob with the value of the db column.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/model/Timestamped.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/model/Timestamped.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.front50.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
 public interface Timestamped {
   /**
    * TODO(rz): Move this method into new Identifiable interface. Has nothing to do with timestamps.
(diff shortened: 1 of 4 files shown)


ajordens:
The same reasoning could be applied to `last_modified_at` as well (and in fact, we were doing something similar back in the S3 days where we'd pull the `last_modified_at` equivalent from object metadata).

Didn't include it in this PR ... but I could.

front50-core/src/main/java/com/netflix/spinnaker/front50/model/Timestamped.java
```diff
@@ -31,4 +33,13 @@
   String getLastModifiedBy();
 
   void setLastModifiedBy(String lastModifiedBy);
+
+  default void setCreatedAt(Long createdAt) {
```
ajordens, line 37: Strangely enough, not all objects in `front50` track created timestamps (from the perspective of the API payloads -- we always have it for SQL).

ajordens:
```
@Override
public Application create(String id, Application application) {
  if (application.getCreateTs() == null) {
    application.setCreateTs(String.valueOf(System.currentTimeMillis()));
  } else {
    application.setUpdateTs(String.valueOf(System.currentTimeMillis()));
  }

  update(id, application);
  return findById(id);
}
```

and

```
@Override
public void update(String id, Application application) {
  application.setName(id);
  super.update(id, application);
}
```

What whatever reason, we don't ever call create().

srekapalli (approved):

PR #1038, opened 2021-06-25
feat(s3): allow enabling payload signing

mattgogerly:
Adds option (disabled by default) to enable payload signing. This is required for Signature V4.

cc @ajordens @german-muzquiz

### DIFF
--- a/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
+++ b/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
@@ -57,8 +57,10 @@ public class S3ClientFactory {
 
     AmazonS3Client client = new AmazonS3Client(awsCredentialsProvider, clientConfiguration);
 
-    S3ClientOptions.Builder clientOptionsBuilder = S3ClientOptions.builder()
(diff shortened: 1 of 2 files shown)


front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
```diff
@@ -37,39 +37,46 @@
...
     }
+
     if (!StringUtils.isEmpty(s3Properties.getSignerOverride())) {
       clientConfiguration.setSignerOverride(s3Properties.getSignerOverride());
     }
 
     AmazonS3Client client = new AmazonS3Client(awsCredentialsProvider, clientConfiguration);
 
+    S3ClientOptions.Builder clientOptionsBuilder =
+        S3ClientOptions.builder().setPayloadSigningEnabled(s3Properties.getPayloadSigning());
```
german-muzquiz, line 61: Just to be extra safe it would be better to check if `s3Properties.getPayloadSigning()` is not null (or convert to `boolean` primitive) before setting this property, in order to avoid `NullPointerException` if the value is ever null.

german-muzquiz (approved):

PR #1044, opened 2021-07-20
fix(api): ignore all Pipeline properties with null value for serialization

mochacat:
#1043 annotated some properties that needed to be excluded if they were null, but we are finding that all of these properties need to be excluded if they are null or there will be null pointer exceptions in Orca.


mochacat:
@danielpeach

danielpeach (approved):

PR #1046, opened 2021-07-23
fix(api): Fix type mismatch with Pipeline boolean properties

mochacat:
Pipeline properties that were migrated to the front50-api package as booleans should actually be strings, since this was the previous behavior (ie true -> "true", false -> "false). This PR reverts these properties back.


edgarulg (approved):
LGTM

PR #1050, opened 2021-08-02
chore(dependencies): Upgrade org.postgresql:postgresql to resolve CVE

j-sandy:
CVE-2020-13692

### DIFF
--- a/front50-sql-postgres/front50-sql-postgres.gradle
+++ b/front50-sql-postgres/front50-sql-postgres.gradle
@@ -1,3 +1,3 @@
 dependencies {
-  runtimeOnly "org.postgresql:postgresql:42.2.8"
+  runtimeOnly "org.postgresql:postgresql:42.2.13" // to fix CVE-2020-13692
 }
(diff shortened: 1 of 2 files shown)


front50-sql-postgres/front50-sql-postgres.gradle
```diff
@@ -1,3 +1,3 @@
 dependencies {
-  runtimeOnly "org.postgresql:postgresql:42.2.8"
+  runtimeOnly "org.postgresql:postgresql:42.2.13" // to fix CVE-2020-13692
```
dbyron-sf, line 2: There seem to be enough places that depend on `org.postgresql:postgresql` that I think it makes sense to move the dependency to [kork](https://github.com/spinnaker/kork/blob/master/spinnaker-dependencies/spinnaker-dependencies.gradle).  For me [this search](https://github.com/search?q=org%3Aspinnaker+org.postgresql%3Apostgresql&type=code) returns clouddriver, fiat, front50, and orca.

j-sandy:
According to recommendation, fix is applied in [Kork.](https://github.com/spinnaker/kork/pull/876) So closing this PR.

PR #1068, opened 2021-09-29
fix(api): revert fix type mismatch regression with Pipeline boolean properties (#1046)

cristhian-castaneda:
The original type for these properties on deck is defined as boolean so we should keep it as is.

Image 2021-09-29 at 3 01 35 PM


edgarulg (approved):
LGTM

edgarulg:
@Mergifyio backport release-1.27.x

PR #1070, opened 2021-10-15
fix(migrations): SpelLoadBalancersMigration runs on every migration

JonGilmore:
`isValid` returns true for `SpelLoadBalancersMigration` and there is no `VALID_UNTIL` set. Because of this, this migration runs regardless of the migration that were planning to run. I imagine the intent was originally to run this once, but in its current configuration, that is not true.

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/SpelLoadBalancersMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/SpelLoadBalancersMigration.java
@@ -2,10 +2,8 @@ package com.netflix.spinnaker.front50.migrations;
 
 import com.netflix.spinnaker.front50.api.model.pipeline.Pipeline;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineDAO;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.time.Clock;
+import java.util.*;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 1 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/SpelLoadBalancersMigration.java
```diff
@@ -2,25 +2,28 @@
...
+import java.util.*;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 
 @Component
 @Slf4j
 public class SpelLoadBalancersMigration implements Migration {
+  // Only valid until October 13th, 2021
+  private static final Date VALID_UNTIL = new GregorianCalendar(2021, 10, 13).getTime();
```
mattgogerly, line 15: Sorry if I'm being dense, what's the significance of this date?

JonGilmore: No apologies necessary, it was just the day before I opened this PR, no significance at all really.

mattgogerly: I'm not familiar with migrations, is the intent that this migration should never run after this PR is merged and released? Is there a way to explicitly enable it after this PR?

Trying to understand if people on older versions of Spinnaker may have issues upgrading if this migration is skipped.

JonGilmore: If you look at some other examples of how Netflix has done migrations, they usually do them 1 of 2 ways - the way that I've done it with a specific date _or_ with a ConditionalOnProperty. The benefit to doing it with a date like this is that you don't have to put up another PR to your config after you're done running the migration to disable it.

>Trying to understand if people on older versions of Spinnaker may have issues upgrading if this migration is skipped.
That's a good question, this migration didn't actually do anything for us because we didn't have any pipelines (out of our thousands that we have) that met this condition. 
Here are the PRs that should give a little more background on what the intent of this migration was for: https://github.com/spinnaker/front50/pull/1025 and https://github.com/spinnaker/deck/pull/8957

JonGilmore: [Here's an example](https://github.com/spinnaker/front50/blob/fe5a9d87a92478eac494f1a511d8ac9ff584297b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/SharedManagedServiceAccountsMigration.java#L43) of an alternative way that migrations have been done

mattgogerly: I think we'll have to wait for @ajordens verdict - I'm not sure if this is a minor issue or a migration everybody needs to do.

PR #1139, opened 2022-07-13
perf(serviceAccounts): Add attribute that allows creating service accounts without running full role syncs

qunayu:
This PR follows on from https://github.com/spinnaker/fiat/pull/961. Main goal is described in that PR. This PR is mainly to demonstrate how that endpoint will be used.

I've added a config value `isDisableRoleSyncWhenSavingServiceAccounts` which can be used to opt-in to this behaviour, making this change quite safe to release.

I will add tests and make sure the build is green once the other PR is merged, as this change depends on a release of `fiat-api` for that new endpoint added. 

Thanks.

### DIFF
--- a/front50-core/src/main/java/com/netflix/spinnaker/front50/ServiceAccountsService.java
+++ b/front50-core/src/main/java/com/netflix/spinnaker/front50/ServiceAccountsService.java
@@ -68,7 +68,11 @@ public class ServiceAccountsService {
 
   public ServiceAccount createServiceAccount(ServiceAccount serviceAccount) {
     ServiceAccount acct = serviceAccountDAO.create(serviceAccount.getId(), serviceAccount);
-    syncUsers(Collections.singletonList(acct));
+    if (fiatConfigurationProperties.isDisableRoleSyncWhenSavingServiceAccounts()) {
+      syncServiceAccount(acct);
+    } else {
+      syncUsers(Collections.singletonList(acct));
+    }
     return acct;
   }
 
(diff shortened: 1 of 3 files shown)


j-sandy:
It seems flag `isDisableRoleSyncWhenSavingServiceAccounts` can be set while building the code and may not be configurable from external configuration. If we can expose this flag, I think that will help us to switch it off and on externally.

qunayu:
@j-sandy  This currently works by toggling `fiat.disableRoleSyncWhenSavingServiceAccounts` in the spring configuration properties, which applies at deploy time. Is that what you meant by switching on/off externally? I'll be happy to use a dynamic feature flag-ish system if spinnaker services are currently supporting it.

j-sandy:
As first step, expected `disableRoleSyncWhenSavingServiceAccounts' to work by defining it in front50-local.yml, verified that it is working. Next logical step should be introducing this flag as part of halyard command to set/unset. I think that can be taken up once this feature get merged.

qunayu:
@j-sandy the property should work when defined within the spring configuration file - `@ConfigurationProperties` is on the class.

jasonmcintosh:
Yeah halyard is kinda dead, and many of these kinds of flags/settings are done at profile vs. halyard layer.  Eventually if this works well, in a release or two, set it as default kinda thing.  Looks like build is failing currently - guessing because Fiat-api isn't updated to support the new API (not dug, just my guess)?

jasonmcintosh (approved):
Looks good thanks for the tests!

qunayu:
Thanks a lot @jasonmcintosh  - could we get this merged in?

PR #1156, opened 2022-08-22
fix(web): Don't treat newly saved pipelines as stale

Nessex:
## WHAT

Fixes `No item found in cache with id of xxxxxxx` errors when copying pipelines.

- When Orca executes the `UpdatePipelinePermissions` task, it creates a random UUID for the new pipeline
- When it executes `SavePipeline` task afterwards, the random UUID is looked up in front50's datastore to validate whether the new pipeline manifest is stale or not. But there is no previous copy of this pipeline as this is the first save

Before this change, this would cause the pipeline save to fail. With this change, the pipeline is saved correctly.

Note: This is only hit when managed service accounts are in-use, because this is the only situation where the pipeline ID will be pre-populated. When you aren't using managed service accounts, there is an existing check for pipeline ID being empty or null.

#### Before
Screen Shot 2022-07-15 at 15 24 06

#### After

Screen Shot 2022-07-15 at 15 24 43

## WHY

These errors cause copying pipelines to fail, when using a very specific configuration of Spinnaker:

- Front50 enabled
- Spinnaker Authz (fiat enabled)
- Using managed service accounts
- Roles field in the pipeline being copied
- Recent version of orca & front50


dbyron-sf:
Please add a test for this.

dbyron-sf:
@mergifyio update

mattgogerly:
Looks like @Nessex hasn't had a chance to look at a test - In the interests of getting this fix in shall we go ahead and merge it?

dbyron-sf:
Let's do it.

dbyron-sf (approved):

mattgogerly:
@Mergifyio update

PR #1172, opened 2022-11-03
fix(pipelines): prevent from creating duplicated pipelines

kkotula:
We've noticed that sometimes spinnaker users end up with duplicated pipelines when using pipelines as a code. In my opinion the logic in PipelineController that creates a pipeline is not safe in a multithreaded environment as we validate pipeline duplicates in line 151, but we actually persist it in line 165. It's likely that two threads running simultaneously will pass validation due to lack of duplicates and then they will persist it.

This simple fix prevents from aforementioned scenario.


ovidiupopa07 (approved):

sergio-quintero (approved):

dogonthehorizon (approved):
Approved, but Chris is going to add a test here before we merge & backport to 1.27, 1.28, and 1.29

dogonthehorizon:
@mergifyio backport release-1.27.x release-1.28.x release-1.29.x

PR #1197, opened 2023-01-11
feat(pipeline/kubernetes): Migrate existing K8s deploy manifest traffic management from redblack to bluegreen

ovidiupopa07:
Migrate existing K8s deploy manifest traffic management from redblack to bluegreen

The migration loads all the k8s pipelines that are currently using red/black traffic management, and it is updating all pipelines to the new terminology blue/green

Pipeline inserted in MySQL using **redblack** strategy
image

After the migration, the strategy will be **bluegreen**
image

image

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
@@ -17,8 +17,6 @@ package com.netflix.spinnaker.front50.migrations;
 
 import com.netflix.spinnaker.front50.api.model.pipeline.Pipeline;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineDAO;
-import java.time.Clock;
-import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -31,8 +29,6 @@ import org.springframework.stereotype.Component;
 @Slf4j
 public class RedBlackToBlueGreenK8sPipelinesMigration implements Migration {
 
-  private static final LocalDate VALID_UNTIL = LocalDate.of(2023, 6, 1);
-
   private final PipelineDAO pipelineDAO;
 
   public RedBlackToBlueGreenK8sPipelinesMigration(PipelineDAO pipelineDAO) {
@@ -41,7 +37,7 @@ public class RedBlackToBlueGreenK8sPipelinesMigration implements Migration {
 
   @Override
   public boolean isValid() {
-    return LocalDate.now(Clock.systemDefaultZone()).isBefore(VALID_UNTIL);
+    return true;
   }
 
   @Override
@@ -60,6 +56,7 @@ public class RedBlackToBlueGreenK8sPipelinesMigration implements Migration {
         .filter(RedBlackToBlueGreenK8sPipelinesMigration::kubernetesProvider)
         .filter(RedBlackToBlueGreenK8sPipelinesMigration::deployManifestStage)
         .map(RedBlackToBlueGreenK8sPipelinesMigration::getTrafficManagement)
+        .filter(Objects::nonNull)
         .filter(RedBlackToBlueGreenK8sPipelinesMigration::trafficManagementEnabled)
         .map(RedBlackToBlueGreenK8sPipelinesMigration::getTrafficManagementOptions)
         .forEach(
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
@@ -22,6 +22,7 @@ import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Predicate;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
@@ -83,6 +84,7 @@ public class RedBlackToBlueGreenK8sPipelinesMigration implements Migration {
             .filter(RedBlackToBlueGreenK8sPipelinesMigration::kubernetesProvider)
             .filter(RedBlackToBlueGreenK8sPipelinesMigration::deployManifestStage)
             .map(RedBlackToBlueGreenK8sPipelinesMigration::getTrafficManagement)
+            .filter(Objects::nonNull)
             .filter(RedBlackToBlueGreenK8sPipelinesMigration::trafficManagementEnabled)
             .map(RedBlackToBlueGreenK8sPipelinesMigration::getTrafficManagementOptions)
             .anyMatch(RedBlackToBlueGreenK8sPipelinesMigration::redBlackStrategy);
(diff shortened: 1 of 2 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
```diff
@@ -0,0 +1,111 @@
...
+    return LocalDate.from(clock.instant()).isBefore(VALID_UNTIL);
+  }
+
+  @Override
+  public void run() {
+    log.info(
+        "Starting the migration of K8s pipelines using RED/BLACK traffic management strategy to BLUE/GREEN");
+    pipelineDAO.all().stream()
+        .filter(RedBlackToBlueGreenK8sPipelinesMigration.pipelineWithRedBlackStrategyPredicate())
+        .forEach(this::migrate);
```
kkotula, line 53: I'd add a log.info() at the end which informs that migration was successfully finished and 123 pipelines were affected, but it's not necessary. I'll leave it to you. 🙃

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Map;
+import java.util.function.Predicate;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class RedBlackToBlueGreenK8sPipelinesMigration implements Migration {
+
+  private static final LocalDate VALID_UNTIL = LocalDate.of(2023, 6, 1);
```
kkotula, line 33: Where is that date coming from? Is it a date of a particular release or something else? It's not wrong, I just want to understand that better.

ovidiupopa07: I chose it - I am still determining what date we should use here.  I am still not sure if instead, we need to make it enabled all the time (since we do not know when customers will use the released version that contains this migration)

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
```diff
@@ -0,0 +1,111 @@
...
+  @Override
+  public boolean isValid() {
+    return LocalDate.from(clock.instant()).isBefore(VALID_UNTIL);
+  }
+
+  @Override
+  public void run() {
+    log.info(
+        "Starting the migration of K8s pipelines using RED/BLACK traffic management strategy to BLUE/GREEN");
+    pipelineDAO.all().stream()
```
kkotula, line 51: I am slightly worried about pulling all pipelines from the database and filtering them in memory. Would it make sense to create designated method to get pipelines that match certain criteria? I can imagine something like pipelineDAO.getAll(String provider) to filter out some pipelines on db level?

I am not saying this is the only way to go. Just sharing a tiny concern.

kkotula (approved):
LGTM! 👏🏻

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
kkotula: Ok, got it! Makes sense!

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
ovidiupopa07: I did it like this since this pattern is used in all other places that interact with the `PipelineDao`. Even the helper method `getPipelinesByApplication` is using the same pattern 
```
public Collection<Pipeline> getPipelinesByApplication(String application, boolean refresh) {
    return all(refresh).stream()
        .filter(
            pipeline ->
                pipeline.getApplication() != null
                    && pipeline.getApplication().equalsIgnoreCase(application))
        .collect(Collectors.toList());
  }
```

I agree this is not the best way to do it, I do not like it either but I think this should be the subject for a different PR

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
ovidiupopa07: Makes sense, I will add the log

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
jasonmcintosh: Actually... this is pipelines.  WE REALLY REALLY don't want PIPELINES In memory.  Those can be LARGE in size.

jasonmcintosh: May not be avoidable with todays system.  Just... thinking ideally this would be batched/paginated at least kinda thing.  Nominally

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
```diff
@@ -0,0 +1,115 @@
...
+        "Migrated pipeline (application: {}, pipelineId: {}, name: {})",
+        pipeline.getApplication(),
+        pipeline.getId(),
+        pipeline.getName());
+  }
+
+  private static Predicate<Pipeline> pipelineWithRedBlackStrategyPredicate() {
+    return pipeline ->
+        pipeline.getStages().stream()
+            .filter(RedBlackToBlueGreenK8sPipelinesMigration::kubernetesProvider)
```
jasonmcintosh, line 83: Ok for now, long term would want AWS as well as any others where the strategy is red/black to keep terminology consistent across spinnaker.

jasonmcintosh (approved):

front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/RedBlackToBlueGreenK8sPipelinesMigration.java
```diff
@@ -0,0 +1,115 @@
...
+      Map<String, Object> trafficManagement) {
+    return (Map<String, Object>) trafficManagement.get("options");
+  }
+
+  private static boolean trafficManagementEnabled(Map<String, Object> trafficManagement) {
+    return Boolean.TRUE.equals(trafficManagement.get("enabled"));
+  }
+
+  private static Map<String, Object> getTrafficManagement(Map<String, Object> stage) {
+    return (Map<String, Object>) stage.get("trafficManagement");
```
jasonmcintosh, line 101: Don't know if trafficManagement can be null, or if that'd break this (I THOUGHT streams filter handled that situation though), but might be worth a test of "no traffic management is set at all

dogonthehorizon (approved):

PR #1211, opened 2023-02-25
feat(migration): ability to disable deleting orphaned objects

mattgogerly:
The `writeOnly` flag is currently unused, so repurposing it to toggle disabling deletion of "orphaned" objects in the target datasource.

Useful when "merging" two disparate instances of Spinnaker together.

### DIFF
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/CompositeStorageServiceConfiguration.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/CompositeStorageServiceConfiguration.kt
@@ -84,5 +84,6 @@ data class StorageServiceMigratorConfigurationProperties(
   var previousClass: String? = null,
   var primaryName: String? = null,
   var previousName: String? = null,
-  var writeOnly: Boolean = false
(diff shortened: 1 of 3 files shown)


dbyron-sf:
With the dynamic config service, it's pretty cheap to add a new flag, yes?  It's not my favorite thing because they're not easily discoverable, and often not well documented...but in this case it might make sense?  Like, it seems a little confusing to use a flag named `writeOnly` to control orphan deletion.

mattgogerly:
> With the dynamic config service, it's pretty cheap to add a new flag, yes? It's not my favorite thing because they're not easily discoverable, and often not well documented...but in this case it might make sense? Like, it seems a little confusing to use a flag named `writeOnly` to control orphan deletion.

Sure. Was going for the path of least resistance but it's an easy change.

dbyron-sf:
Of course this is where I ask for an automated test, but the change looks OK to me.

mattgogerly:
> Of course this is where I ask for an automated test, but the change looks OK to me.

I was one step ahead of you until I forgot to stage the test class.

dbyron-sf (approved):

PR #1249, opened 2023-04-15
feat(core + sql): Optimize cache refresh

dbyron-sf:
This PR adds three new config flags to [each object type under service-storage](https://github.com/spinnaker/front50/blob/f6452ec07f207246b3ee940c49d31a115886d4a3/front50-core/src/main/java/com/netflix/spinnaker/front50/config/StorageServiceConfigurationProperties.java#L9-L20).

One is fairly mechanical:
* cacheHealthCheckTimeoutSeconds: previously hard-coded to 90

The other two are performance improvements.  They default to false.
* synchronizeCacheRefresh: When true, if multiple threads attempt to refresh the cache in StorageServiceSupport simultaneously, only one actually does the refresh.  The others wait until it's complete.  This reduces load on the data store.

* optimizeCacheRefreshes: When true, for objects that support versioning, cache refreshes only query the data store for objects modified (or deleted) since the last refresh.  This is only implemented for sql.

I don't have before/after metrics for this, but we've been running this code with a mysql data store for about 2 years configured as follows:
```
storage-service:
  application:
    optimizeCacheRefreshes: true
    synchronizeCacheRefresh: true
  pipeline:
    optimizeCacheRefreshes: true
    synchronizeCacheRefresh: true
```
and front50 database load hasn't been on the radar as an optimization target.  We've got a whole bunch of pipelines.

### DIFF
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
@@ -101,7 +101,7 @@ class SqlStorageService(
           .from(definitionsByType[objectType]!!.tableName)
           .where(
             field("id", String::class.java).eq(objectKey).and(
-              DSL.field("is_deleted", Boolean::class.java).eq(false)
+              field("is_deleted", Boolean::class.java).eq(false)
             )
           )
           .fetchOne()
@@ -133,7 +133,7 @@ class SqlStorageService(
                 .from(definitionsByType[objectType]!!.tableName)
                 .where(
                   field("id", String::class.java).`in`(keys).and(
-                    DSL.field("is_deleted", Boolean::class.java).eq(false)
+                    field("is_deleted", Boolean::class.java).eq(false)
                   )
                 )
                 .fetch()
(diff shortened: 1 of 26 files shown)


front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -164,20 +167,76 @@ class SqlStorageService(
...
+    val notDeletedKey = "not_deleted"
+
+    val resultMap = mapOf<String, MutableList<T>>(
+      deletedKey to mutableListOf(),
+      notDeletedKey to mutableListOf()
+    )
+
+    val tableSupportsSoftDeletes = definitionsByType[objectType]!!.supportsHistory
+
+    val fieldsToFetch = mutableListOf<SelectFieldOrAsterisk>(field("body", String::class.java))
```
jvz, line 190: Sort of a nitpick, but I've recently started using [this `SelectQuery` API](https://www.jooq.org/javadoc/3.14.x/org.jooq/org/jooq/SelectQuery.html) for assembling dynamic queries like this. This might simplify things for you. See also [`InsertQuery`](https://www.jooq.org/javadoc/3.14.x/org.jooq/org/jooq/InsertQuery.html) which also abstracts away some MySQL/Postgres differences in the plain DSL.

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -164,20 +167,76 @@ class SqlStorageService(
...
+    if (tableSupportsSoftDeletes) {
+      fieldsToFetch.add(field("is_deleted", Boolean::class.java))
+    }
+
+    val timeToLoadObjects = measureTimeMillis {
+      val result: Result<Record> = jooq.withRetry(sqlRetryProperties.reads) { ctx ->
+        ctx
+          .select(fieldsToFetch)
+          .from(definitionsByType[objectType]!!.tableName)
+          .where(field("last_modified_at", Long::class.java).greaterThan(lastModifiedThreshold))
```
jvz, line 200: Think it's worth adding an index on this column?

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
dbyron-sf: These sound like good improvements.  Any chance we can leave them for a followup PR?  We've been running this code as is for a long time and I'd love to avoid another round of testing.

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
dbyron-sf: Done in https://github.com/spinnaker/front50/pull/1256.

PR #1251, opened 2023-04-19
feat(pipelines): add new GET /pipelines/triggeredBy/{id:.+}/{status} endpoint

dbyron-sf:
to return all pipelines triggered when the given pipeline configuration id completes with the given status.

Initially used by https://github.com/spinnaker/orca/pull/4448.


front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Trigger.java
```diff
@@ -28,6 +32,76 @@ protected Map<String, Object> delegate() {
     return backing;
   }
 
+  public Boolean getEnabled() {
+    Object enabled = backing.get("enabled");
+    if (enabled instanceof Boolean) {
+      return (Boolean) enabled;
+    }
+
+    // orca's DependentPipelineExecutionListener.groovy currently (18-apr-23)
```
dbyron-sf, line 41: See [here](https://github.com/spinnaker/orca/blob/da3cf93de4f003a91d3ddd01c1996d0a8961e58f/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy#L97) for the orca code.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -90,6 +92,48 @@ public Collection<Pipeline> list(
...
+              id,
+              status,
+              retval ? "matches" : "does not match",
+              trigger.toString());
+          return retval;
+        };
+
+    // Return all pipelines with at least one trigger defined for the given id
+    // and status.
+    Predicate<Pipeline> pipelinePredicate =
```
dbyron-sf, line 131: This is supposed to be equivalent to [this logic in orca](https://github.com/spinnaker/orca/blob/da3cf93de4f003a91d3ddd01c1996d0a8961e58f/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy#L93-L101), except for the difference highlighted [here](https://github.com/spinnaker/front50/pull/1251/files#diff-ef16150a5e2666115561825274021544fe6e234f34b5cd5f7bd29c67469afa49R41).

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Trigger.java
```diff
@@ -28,6 +32,76 @@ protected Map<String, Object> delegate() {
...
+      // This allows not only true/True/TRUE but also some values that groovy
+      // considers truthy, e.g. tRue, tRUe, etc.
+      return Boolean.valueOf(((String) enabled).toLowerCase());
+    }
+
+    if (enabled != null) {
+      // It's tempting to warn if enabled is some other type, as it's likely a
+      // pipeline bug.  There's not enough context here to help find the
+      // offending pipeline though, so let's leave it as debug.
+      log.debug(
```
jasonmcintosh, line 82: Should be a WARN vs. debug?  Possibly including the name of the pipeline or some other details be ideal as well.

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Trigger.java
```diff
@@ -28,6 +32,76 @@ protected Map<String, Object> delegate() {
...
+  }
+
+  public List<String> getStatus() {
+    Object status = backing.get("status");
+    if (status instanceof List) {
+      return List.copyOf((List<String>) status);
+    }
+
+    if (status != null) {
+      log.debug("status: '{}' is a {}, not a List", status, status.getClass().getSimpleName());
```
jasonmcintosh, line 100: Same on the debug... this seems problematic to hit.

jasonmcintosh (approved):

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Trigger.java
dbyron-sf: Right...the comment is trying to explain this.  I can go back to warn, but there isn't really any other context here.  If we're lucky the log messages would have e.g. X-SPINNAKER-REQUEST-ID.

jasonmcintosh: yeah I THOUGHT there was context in the backing object :( IN THEORY X-Spinnaker-Applicaiton or similar headers?

dbyron-sf: https://github.com/spinnaker/front50/pull/1251/commits/f947b03ef566b04ea64030c85d4eeb2313b501f3

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Trigger.java
dbyron-sf: Changed to warning in https://github.com/spinnaker/front50/pull/1251/commits/f947b03ef566b04ea64030c85d4eeb2313b501f3

PR #1252, opened 2023-04-20
feat(web): add optional query params to the GET /pipelines endpoint

dbyron-sf:
so it's possible for consumers of pipelines (e.g. echo) to get only the information they need.

enabledPipelines: boolean -- only return enabled/disabled pipelines
enabledTriggers: boolean -- only return pipelines with at least one enabled trigger
triggerTypes: only return pipelines with a trigger whose type is present in this comma-separated string of trigger types

for example GET /pipelines?enabledPipelines=true&enabledTriggers=true&triggerTypes=foo,bar

which, with correct URL-encoding is:

GET /pipelines?enabledPipelines=true&enabledTriggers=true&triggerTypes=foo%2Cbar

Note that if both enabledTriggers and triggerTypes are specified, that a single trigger must satisfy both conditions for front50 to return its pipeline.

Also add a GET /pipelines/{application}/name/{name} endpoint which queries for a pipeline by application and name. It responds with 404 if no such pipeline is found.

echo uses this endpoint in https://github.com/spinnaker/echo/pull/1292 for manual triggers by application and name, when its cache doesn't contain the triggered pipeline.

### DIFF
--- a/front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSpec.groovy
+++ b/front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSpec.groovy
@@ -254,6 +254,14 @@ class PipelineControllerSpec extends Specification {
     }
 
     @Override
+    Pipeline getPipelineByName(String application, String pipelineName, boolean refresh) {
+      map.values().stream()
+        .filter({ p -> p.getApplication().equalsIgnoreCase(application) })
+        .filter({ p -> p.getName().equalsIgnoreCase(pipelineName) })
+        .collect(Collectors.toList())
+    }
+
+    @Override
     Pipeline findById(String id) throws NotFoundException {
       def get = map.get(id)
       if (get == null) {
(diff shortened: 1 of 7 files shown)


front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSpec.groovy
```diff
@@ -253,6 +253,14 @@ class PipelineControllerSpec extends Specification {
         .collect(Collectors.toList())
     }
 
+    @Override
+    Pipeline getPipelineByName(String application, String pipelineName, boolean refresh) {
+      map.values().stream()
+        .filter({ p -> p.getApplication().equalsIgnoreCase(application) })
+        .filter({ p -> p.getName().equalsIgnoreCase(pipelineName) })
```
jasonmcintosh, line 260: YAY a case where we ignore case :)

jasonmcintosh (approved):

front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSpec.groovy
dbyron-sf: Trying to follow what similar methods do above ...

PR #1275, opened 2023-06-30
feat(sql): add configuration property sql.healthIntervalMillis

dbyron-sf:
that controls how often the health endpoint refreshes information when using sql.  It defaults to 30 seconds to match the behavior before this change.

Information about cache status for each object type (e.g. pipelines, applications, etc.) contributes to the response from front50's /health endpoint.  This change makes it possible to reduce the likelihood that the cache has refreshed, but the response from the /health endpoint doesn't reflect that.

### DIFF
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/Front50SqlProperties.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/config/Front50SqlProperties.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2023 Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


front50-sql/src/main/kotlin/com/netflix/spinnaker/config/Front50SqlProperties.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Note that kork's SqlProperties class also uses the "sql" prefix.
+ *
+ * @param healthIntervalMillis The period to refresh health information (e.g. in the health endpoint).
+ */
+@ConfigurationProperties("sql")
+class Front50SqlProperties {
+  /**
+   * How frequently to refresh health information (e.g. for the health endpoint).
+   */
+  var healthIntervalMillis: Long = Duration.ofSeconds(30).toMillis()
```
jvz, line 31: Nit: this _could_ be a `Duration` and converted to other duration units on demand.

jvz (approved):

front50-sql/src/main/kotlin/com/netflix/spinnaker/config/Front50SqlProperties.kt
dbyron-sf: Yeah.  I still need to wrap my head around what that means for variable names and how config files look to do it that way.

PR #1279, opened 2023-06-30
fix(core): skip existing items with null ids in StorageServiceSupport.fetchAllItemsOptimized

dbyron-sf:
Even though it seems like the existing filters on items coming back from the database would make this unnecessary, given that existingItems starts empty and, from what I can tell, only contains (filtered) items from the database.

We've seen this happen in production though.  With any luck the logging will help us figure out why.


jvz (approved):

dbyron-sf:
@mergifyio backport release-1.31.x

PR #1452, opened 2024-04-04
fix(web): restore check for regenerateCronTriggerIds in PipelineController.save

dbyron-sf:
https://github.com/spinnaker/front50/pull/1035/files#diff-9b514be177faf5444c86a88ab6bb9e6a0add032bfa67862bc8e33b17c4bb9cc9L159 removed it, but [orca's SavePipelineTask](https://github.com/spinnaker/orca/blob/2a2110d849bc50b2497e40554fead7a67abcf8ee/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java#L95) still sets it.

with some little refactors along the way.

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
@@ -252,11 +252,9 @@ public class PipelineController {
     if (Strings.isNullOrEmpty(pipeline.getId())
         || (boolean) pipeline.getAny().getOrDefault("regenerateCronTriggerIds", false)) {
       // ensure that cron triggers are assigned a unique identifier for new pipelines
-      List<Trigger> triggers = pipeline.getTriggers();
-      triggers.stream()
+      pipeline.getTriggers().stream()
           .filter(it -> "cron".equals(it.getType()))
           .forEach(it -> it.put("id", UUID.randomUUID().toString()));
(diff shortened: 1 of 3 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -249,13 +249,12 @@ public synchronized Pipeline save(
     pipeline.setName(pipeline.getName().trim());
     pipeline = ensureCronTriggersHaveIdentifier(pipeline);
 
-    if (Strings.isNullOrEmpty(pipeline.getId())) {
+    if (Strings.isNullOrEmpty(pipeline.getId())
+        || (boolean) pipeline.getAny().getOrDefault("regenerateCronTriggerIds", false)) {
       // ensure that cron triggers are assigned a unique identifier for new pipelines
-      List<Trigger> triggers = pipeline.getTriggers();
-      triggers.stream()
+      pipeline.getTriggers().stream()
```
dbyron-sf, line 255: https://github.com/spinnaker/front50/pull/987/files#r515405296 is no longer valid after https://github.com/spinnaker/front50/pull/1035/files#diff-0b2bc300fd3965c64ba4184955384aa8609cace04e1c015b31ef4a83552e53d1R36.  There's no need to get/modify/set anymore.

jasonmcintosh (approved):

PR #1457, opened 2024-04-18
refactor(web): change ensureCronTriggersHaveIdentifier to return void

dbyron-sf:
and use an additional filter to clean up the code.

With [pipeline.getTriggers()](https://github.com/spinnaker/front50/pull/1035/files#diff-0b2bc300fd3965c64ba4184955384aa8609cace04e1c015b31ef4a83552e53d1R36) implemented as an accessor, ensureCronTriggersHaveIdentifier was already mutating the pipeline.  Let's make that more obvious by making it return void and adding some javadoc.


nemesisOsorio (approved):
LGTM

PR #1458, opened 2024-04-19
feat(migrations): Support for migrations defined in plugins

jervi:
The current implementation of the MigrationRunner is initialized before any migrations in plugins are initialized, and they are therefore not included by the runner. Even if the plugin migrations are defined using `@ExposeToApp`, the beans are never picked up by the MigrationRunner bean. This PR changes the MigrationRunner to look up Migration beans directly from the applicationContext on every run, and will also delay the first run with 10 seconds in order to give some time for plugins to be initialized.

### DIFF
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
@@ -16,14 +16,14 @@
 
 package com.netflix.spinnaker.front50.migrations;
 
+import static net.logstash.logback.argument.StructuredArguments.value;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.ApplicationContext;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
-
-import static net.logstash.logback.argument.StructuredArguments.value;
 
 /**
  * Migration runner runs all the registered migrations as scheduled. By default migration runner
--- a/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
+++ b/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
@@ -16,15 +16,14 @@
 
 package com.netflix.spinnaker.front50.migrations;
 
-import static net.logstash.logback.argument.StructuredArguments.value;
-
-import java.util.Collection;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations/MigrationRunner.java
```diff
@@ -38,17 +37,25 @@
...
+  public MigrationRunner(ApplicationContext applicationContext) {
+    this.applicationContext = applicationContext;
+  }
 
   /** Run migrations every 8hrs. */
-  @Scheduled(fixedDelay = 28800000)
+  @Scheduled(fixedDelay = 28800000, initialDelay = 10000)
   void run() {
-    migrations.stream()
+    applicationContext.getBeansOfType(Migration.class).values().stream()
```
jasonmcintosh, line 49: A concern is this would reload the migrations list on EVERY run - this may NOT be as efficient.  I'm not against this, just a call out that it's a deeper load on the calls as a result.  


ALSO... tests please!  I know no PREVIOUS tests, just a simple test at least on the valid logic would be useful.

jasonmcintosh: ALSO the delay - does it matter since this will rerun and next run pick up those plugin registered migrations?

jasonmcintosh: OR at least possibly a config value to set the delay with a sane default -  so plugins/loads that take longer than the 10 seconds will handle this.

jervi: It's annoying to wait 8 hours to see if your migration worked as expected, so I'd say it should be there. But I can make it configurable. Maybe I can make the 8 hours configurable as well then

jervi: I don't think looking up 8 beans from the application context every 8 hour is any concern

jervi: > ALSO... tests please!

Added! Not sure how helpful it is, I feel like I'm mostly testing Spring. But it works.

jasonmcintosh: Yeah it's... mostly testing spring, but really making sure to test the enabled/disabled migrator configs ;) Thanks for that!  AND the config delay - 8 hours would be horrible you're right :) Being able to adjust to say 5 minutes though should resolve a situation I know on some plugins delayed loading...

jasonmcintosh (approved):

PR #1459, opened 2024-04-23
fix(web): Retrieve dependent pipelines correctly

nandak522:
### Issue Summary
This fixes the bug regarding `/v2/pipelineTemplates/<pipeline-template-id>/dependentPipelines` API.
Original Issue: https://github.com/spinnaker/spinnaker/issues/6472

### Environment
Spinnaker: `1.33.0+`
Front50: `us-docker.pkg.dev/spinnaker-community/docker/front50:2.31.0`

### Description
Given a `templateId`, the `getDependentConfigs` method called by `listDependentPipelines` method, was returning `null` for pipeline configuration, causing the entire API to return `[]` even if a specific pipeline template had dependent pipelines.. This seems to be because of 👇 change.
```java
import com.netflix.spinnaker.front50.model.pipeline.TemplateConfiguration;
TemplateConfiguration v1Config = objectMapper.convertValue(templatedPipeline.getConfig(), TemplateConfiguration.class);

// `templatedPipeline.getConfig()` above was returning `null`
```
This 👆 I believe is meant for `v1` templates. But for `v2` templates I believe it has to be 👇 :
```java
import com.netflix.spinnaker.front50.model.pipeline.V2TemplateConfiguration;

V2TemplateConfiguration config = objectMapper.convertValue(templatedPipeline, V2TemplateConfiguration.class);
```

Here are the swagger-ui screenshots showing before and after the fix 👇 
| Before | After |
|--------|--------|
| Before | After|

Here are the cURL responses showing before and after the fix 👇 
#### Before
```sh
curl -X GET "http://localhost:30084/v2/pipelineTemplates/9d92ff72-2bbf-43a3-bd57-23c221387051/dependents" -H "accept: */*"
[]
```

#### After
```sh
curl -X GET "http://localhost:30084/v2/pipelineTemplates/9d92ff72-2bbf-43a3-bd57-23c221387051/dependents" -H "accept: */*"
[{"application":"products","pipelineConfigId":"9aef1986-48cf-44ce-859c-97b3042388db","pipelineName":"[Mother Pipeline] Global Deployment Express"}]
```

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
@@ -31,7 +31,7 @@ import com.netflix.spinnaker.front50.exceptions.InvalidRequestException;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineDAO;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplate;
 import com.netflix.spinnaker.front50.model.pipeline.PipelineTemplateDAO;
-import com.netflix.spinnaker.front50.model.pipeline.TemplateConfiguration;
+import com.netflix.spinnaker.front50.model.pipeline.V2TemplateConfiguration;
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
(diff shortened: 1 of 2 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
```diff
@@ -186,22 +186,23 @@ List<String> getDependentConfigs(String templateId) {
...
-                TemplateConfiguration config =
-                    objectMapper.convertValue(
-                        templatedPipeline.getConfig(), TemplateConfiguration.class);
 
-                source = config.getPipeline().getTemplate().getSource();
+                V2TemplateConfiguration config =
+                    objectMapper.convertValue(templatedPipeline, V2TemplateConfiguration.class);
+                source = config.getTemplate().getReference();
               } catch (Exception e) {
+                e.printStackTrace();
```
dbyron-sf, line 201: We don't typically print to standard out in shipping code.  If there's something to log here, please do.

dbyron-sf:
Can you add some automated tests that demonstrate the bug?  I'd also like to see a more specific commit message that mentions what was broken.  Thanks.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java
nandak522: This is removed.

nandak522:
@dbyron-sf I have added the tests. And reworded the commit message. Please take a look.

nandak522:
> Can you add some automated tests that demonstrate the bug? 
Here it is: @dbyron-sf : https://github.com/spinnaker/front50/pull/1459/commits/1402d7b651ba94cc2fe80bb533f861ab9a8629b8#diff-f2c1cc675d8e520ff0541b397bd0d390483bcf197b0576fe2b5976a8db2de02a

dbyron-sf (approved):

PR #1483, opened 2024-07-19
feat(batchUpdate): enhance batch update functionality

kirangodishala:
This PR adds the following functionality to pipeline batch update operation :
- refactor bulk storage operation to make it atomic
- add duplicate pipeline check to /pipelines POST handler method (i.e. save individual pipeline) and  /pipelines/{id} PUT handler method (i.e. update individual pipeline)
- update /pipelines/batchUpdate POST handler method to address deserialization issues and add some useful log statements
- adds configurable controls to decide whether cache should be refreshed while checking for duplicate pipelines 
```yaml
controller:
   pipeline:
      save:
         refreshCacheOnDuplicatesCheck: false // default is true 
```
- Add pipeline validations to validate all the deserialized pipelines
-  batchUpdate now responds with a status of succeeded and failed pipelines info. The response will be a map containing information in the following format:
```
[
"successful_pipelines_count"  : <int>,
"successful_pipelines"        : <List<String>>,
"failed_pipelines_count"      : <int>,
"failed_pipelines"            : <List<Map<String, Object>>> 
]
```
- add staleCheck to batchUpdate
- adjust permissions to batchUpdate 
   - before: isAdmin, 
   - now: verifies application write permission

### DIFF
--- a/front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
+++ b/front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
@@ -25,7 +25,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 
-@EqualsAndHashCode
+@EqualsAndHashCode(of = {"id", "name", "application"})
 public class Pipeline implements Timestamped {
 
   public static final String TYPE_TEMPLATED = "templatedPipeline";
--- a/front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
+++ b/front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
@@ -21,9 +21,11 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 
+@EqualsAndHashCode
 public class Pipeline implements Timestamped {
 
   public static final String TYPE_TEMPLATED = "templatedPipeline";
--- a/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
+++ b/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
@@ -258,63 +258,58 @@ class SqlStorageService(
   }
 
   override fun <T : Timestamped> storeObjects(objectType: ObjectType, allItems: Collection<T>) {
-    // using a lower `chunkSize` to avoid exceeding default packet size limits.
-    allItems.chunked(100).forEach { items ->
-      try {
-        withPool(poolName) {
-          jooq.transactional(sqlRetryProperties.transactions) { ctx ->
+    withPool(poolName) {
+      jooq.transactional(sqlRetryProperties.transactions) { ctx ->
+        // using a lower `chunkSize` to avoid exceeding default packet size limits.
+        allItems.chunked(100).forEach { items ->
+          try {
+            ctx.batch(
+              items.map { item ->
+                val insertPairs = definitionsByType[objectType]!!.getInsertPairs(
+                  objectMapper, item.id.toLowerCase(), item
+                )
+                val updatePairs = definitionsByType[objectType]!!.getUpdatePairs(insertPairs)
+
+                ctx.insertInto(
+                  table(definitionsByType[objectType]!!.tableName),
+                  *insertPairs.keys.map { field(it) }.toTypedArray()
+                )
+                  .values(insertPairs.values)
+                  .onConflict(field("id", String::class.java))
+                  .doUpdate()
+                  .set(updatePairs.mapKeys { field(it.key) })
+              }
+            ).execute()
+          } catch (e: SQLDialectNotSupportedException) {
+            for (item in items) {
+              storeSingleObject(objectType, item.id.toLowerCase(), item)
+            }
+          }
+
+          if (definitionsByType[objectType]!!.supportsHistory) {
             try {
               ctx.batch(
                 items.map { item ->
-                  val insertPairs = definitionsByType[objectType]!!.getInsertPairs(
-                    objectMapper, item.id.toLowerCase(), item
+                  val historyPairs = definitionsByType[objectType]!!.getHistoryPairs(
+                    objectMapper, clock, item.id.toLowerCase(), item
                   )
-                  val updatePairs = definitionsByType[objectType]!!.getUpdatePairs(insertPairs)
 
-                  ctx.insertInto(
-                    table(definitionsByType[objectType]!!.tableName),
-                    *insertPairs.keys.map { field(it) }.toTypedArray()
-                  )
-                    .values(insertPairs.values)
-                    .onConflict(field("id", String::class.java))
-                    .doUpdate()
-                    .set(updatePairs.mapKeys { field(it.key) })
+                  ctx
+                    .insertInto(
+                      table(definitionsByType[objectType]!!.historyTableName),
+                      *historyPairs.keys.map { field(it) }.toTypedArray()
+                    )
+                    .values(historyPairs.values)
+                    .onDuplicateKeyIgnore()
                 }
               ).execute()
             } catch (e: SQLDialectNotSupportedException) {
               for (item in items) {
-                storeSingleObject(objectType, item.id.toLowerCase(), item)
-              }
-            }
-
-            if (definitionsByType[objectType]!!.supportsHistory) {
-              try {
-                ctx.batch(
-                  items.map { item ->
-                    val historyPairs = definitionsByType[objectType]!!.getHistoryPairs(
-                      objectMapper, clock, item.id.toLowerCase(), item
-                    )
-
-                    ctx
-                      .insertInto(
-                        table(definitionsByType[objectType]!!.historyTableName),
-                        *historyPairs.keys.map { field(it) }.toTypedArray()
-                      )
-                      .values(historyPairs.values)
-                      .onDuplicateKeyIgnore()
-                  }
-                ).execute()
-              } catch (e: SQLDialectNotSupportedException) {
-                for (item in items) {
-                  storeSingleObjectHistory(objectType, item.id.toLowerCase(), item)
-                }
+                storeSingleObjectHistory(objectType, item.id.toLowerCase(), item)
               }
             }
           }
         }
-      } catch (e: Exception) {
-        log.error("Unable to store objects (objectType: {}, objectKeys: {})", objectType, items.map { it.id })
-        throw e
       }
     }
   }
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
@@ -19,8 +19,10 @@ import static com.netflix.spinnaker.front50.api.model.pipeline.Pipeline.TYPE_TEM
 import static com.netflix.spinnaker.front50.model.pipeline.TemplateConfiguration.TemplateSource.SPINNAKER_PREFIX;
 import static java.lang.String.format;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
 import com.netflix.spinnaker.front50.ServiceAccountsService;
 import com.netflix.spinnaker.front50.api.model.pipeline.Pipeline;
 import com.netflix.spinnaker.front50.api.model.pipeline.Trigger;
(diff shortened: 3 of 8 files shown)


dbyron-sf:
For the record, https://github.com/spinnaker/front50/pull/988 was a previous attempt at something similar.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
+
+    long bulkImportStartTime = System.currentTimeMillis();
+    log.debug("Bulk importing the following pipelines: {}", pipelinesToSave);
+    pipelineDAO.bulkImport(pipelinesToSave);
+    log.debug(
+        "Bulk imported {} pipelines successfully in {}ms",
+        pipelinesToSave.size(),
+        System.currentTimeMillis() - bulkImportStartTime);
+
+    List<String> savedPipelines =
```
link108, line 342: in the failed pipelines, you're storing more info about each pipeline, eg application, id, etc. does it make sense to add some of that information to the saved pipelines as well?

dbyron-sf: I see where we store error information in failed pipelines, but not other stuff.  Can you provide a pointer?

kirangodishala: The whole failed pipeline(as a map) gets added to the list of failed pipelines in response. [Here](https://github.com/spinnaker/front50/blob/894299a067373ad289764ab0711e038ffb61ab09/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java#L424-L434) and [here](https://github.com/spinnaker/front50/blob/894299a067373ad289764ab0711e038ffb61ab09/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java#L329-L332) is the code for that. Additionally`errorMsg` is put into the failed pipeline map.

kirangodishala: @link108 - Given hundreds/thousands of pipelines, the idea I think is to let the user know more about failed pipelines and keep the info about succeeded ones to minimum. But if adding a couple of fields (like id, application) provides more value, we can do that.

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -258,63 +258,58 @@ class SqlStorageService(
...
-              } catch (e: SQLDialectNotSupportedException) {
-                for (item in items) {
-                  storeSingleObjectHistory(objectType, item.id.toLowerCase(), item)
-                }
+                storeSingleObjectHistory(objectType, item.id.toLowerCase(), item)
               }
             }
           }
         }
-      } catch (e: Exception) {
```
jasonmcintosh, line 315: I doubt this happens much - but any reason to lose this try/catch here?

jasonmcintosh: Thinking TransientDaoException case where DB is failing or similar...

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
```diff
@@ -258,63 +258,58 @@ class SqlStorageService(
   }
 
   override fun <T : Timestamped> storeObjects(objectType: ObjectType, allItems: Collection<T>) {
-    // using a lower `chunkSize` to avoid exceeding default packet size limits.
-    allItems.chunked(100).forEach { items ->
-      try {
-        withPool(poolName) {
-          jooq.transactional(sqlRetryProperties.transactions) { ctx ->
+    withPool(poolName) {
```
jasonmcintosh, line 261: With this flip, I think the entire batch fails now vs. only a chunk of it - intentional?

jasonmcintosh: Guess that's partly the point of this PR - just wondering if that's a good thing?

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
   @RequestMapping(value = "", method = RequestMethod.POST)
   public synchronized Pipeline save(
       @RequestBody Pipeline pipeline,
       @RequestParam(value = "staleCheck", required = false, defaultValue = "false")
           Boolean staleCheck) {
 
-    validatePipeline(pipeline, staleCheck);
+    long saveStartTime = System.currentTimeMillis();
+    log.info(
+        "Received request to save pipeline {} in application {}",
```
jasonmcintosh, line 272: NIT:  Info?  We get a LOT of these in some cases and that'll add a lot of logs.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
-          .forEach(it -> it.put("id", UUID.randomUUID().toString()));
-    }
+    log.debug(
+        "Successfully validated pipeline {} in {}ms",
+        pipeline.getName(),
+        System.currentTimeMillis() - validationStartTime);
 
-    return pipelineDAO.create(pipeline.getId(), pipeline);
+    Pipeline savedPipeline = pipelineDAO.create(pipeline.getId(), pipeline);
+    log.info(
```
jasonmcintosh, line 288: Same nit:  probably should be debug level vs. info level.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
+    long saveStartTime = System.currentTimeMillis();
+    log.info(
+        "Received request to save pipeline {} in application {}",
+        pipeline.getName(),
+        pipeline.getApplication());
 
+    log.debug("Running validation before saving pipeline {}", pipeline.getName());
+    long validationStartTime = System.currentTimeMillis();
+    validatePipeline(pipeline, staleCheck);
+    checkForDuplicatePipeline(
```
jasonmcintosh, line 279: Minor:  Should this be part of validation? It's done on both save & batch & would reduce the duplicate call I'd think...

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
jasonmcintosh: Can ignore this mostly just curious philosophically which is better :)

front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/model/SqlStorageService.kt
jasonmcintosh: SHORT answer it'd be nice if these were caught & re-raised as Spinnaker exception objects

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
+        pipeline.getApplication(), pipeline.getName().trim(), pipeline.getId());
     pipeline.setName(pipeline.getName().trim());
```
jasonmcintosh, line 281: This is being done in validatePipeline so don't think this is needed here anymore.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -355,18 +492,29 @@ private void validatePipeline(final Pipeline pipeline, Boolean staleCheck) {
...
+    // Run stale pipeline definition check
     if (staleCheck
         && !Strings.isNullOrEmpty(pipeline.getId())
         && pipeline.getLastModified() != null) {
       checkForStalePipeline(pipeline, errors);
     }
 
+    // Run other pre-configured validators
+    pipelineValidators.forEach(it -> it.validate(pipeline, errors));
+
```
jasonmcintosh, line 517: Honestly a LOT of the above operations LIKE the cron ids could be moved to pipelineValidators & shrink a lot of this code down :) and simplify it! Future enhancement maybe...

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
```diff
@@ -21,9 +21,11 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 
+@EqualsAndHashCode
```
jasonmcintosh, line 28: CAUTION:  This is... going to do a deep comparison on all fields as a result.  It'd be better to do it on explicit fields (aka id) or similar.  I could see this leading to some performance issues as a result.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
kirangodishala: As this is an important info which unambiguously confirms the pipeline save, kept it under info so that it's more readable and useful for monitoring.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
kirangodishala: It's not called under batch. The way duplicates are checked is different for save and batch hence the common code is moved to validation and this method is kept outside.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -237,33 +258,106 @@ public Pipeline getByApplicationAndName(
...
+    // The right side of the pair holds failed pipelines.  This needs to be
+    // List<Map<String, Object>> as opposed to List<Pipeline> since some
+    // elements of pipelineJson may fail to deserialize into Pipeline objects.
+    ImmutablePair<List<Pipeline>, List<Map<String, Object>>> deserializedPipelines =
+        deSerializePipelines(pipelinesJson);
+    List<Pipeline> pipelines = deserializedPipelines.getLeft();
+    List<Map<String, Object>> failedPipelines = deserializedPipelines.getRight();
+
+    log.info(
```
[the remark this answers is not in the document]
kirangodishala: This log occurs once for the entire batch operation so I am hoping it comes handy while monitoring.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
kirangodishala: agreed. Will change.

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
kirangodishala: would you think the three fields - id, name and application be better?

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
jasonmcintosh: Monitoring via logs isn't great monitoring.  That's really more a category for metrics/event emission behavior.  Nominally an echo event (e.g. when using Orca TASKS this is automatically emitted via an echo event for example... since this bypasses that, that kind of thing becomes MUCH more important).

front50-api/src/main/java/com/netflix/spinnaker/front50/api/model/pipeline/Pipeline.java
jasonmcintosh: yeah :) We don't want deep introspection as this would have some potentially nasty impacts elsewhere.  Frankly ID is probably ALL that's needed as that's supposed to be unique, but I'd be ok with name/application as well if we really want.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
dbyron-sf: Can we leave that for a subsequent PR?

jasonmcintosh: I'm ok with that - biggest concern is audit log behavior and NOT capturing these via normal audit logs.  This results in missed compliance possibilities.  A separate PR to make sure these events ARE auditied/sent through that would work.

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
jasonmcintosh: Ah k then I mis-read, thought it was duplicated between the two (the check for duplicate)

kirangodishala:
@link108 @jasonmcintosh - would you approve the PR?

link108 (approved):

PR #1520, opened 2024-12-19
feat(web): Exclude Disabled pipelines query parameter

christosarvanitis:
Optionally exclude the Disabled pipelines of an application. This can be opt-in by the Orca PR https://github.com/spinnaker/orca/pull/4819

### DIFF
--- a/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
+++ b/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
@@ -202,11 +202,36 @@ public class PipelineController {
   public List<Pipeline> listByApplication(
       @PathVariable(value = "application") String application,
       @RequestParam(value = "pipelineNameFilter", required = false) String pipelineNameFilter,
-      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh) {
+      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh,
+      @RequestParam(required = false, value = "enabledPipelines", defaultValue = "false")
+          Boolean enabledPipelines) {
     List<Pipeline> pipelines =
         new ArrayList<>(
             pipelineDAO.getPipelinesByApplication(application, pipelineNameFilter, refresh));
 
+    if (!enabledPipelines) {
+      return sortPipelines(pipelines);
+    }
+
+    Predicate<Pipeline> pipelinePredicate =
+        pipeline -> {
+          // pipeline.getDisabled may be null, so check that before comparing.  If
+          // pipeline.getDisabled is null, the pipeline is enabled.
+          boolean pipelineEnabled =
+              (pipeline.getDisabled() == null) || (pipeline.getDisabled() == false);
+
+          return ((enabledPipelines == null) || (pipelineEnabled == enabledPipelines));
+        };
+
+    List<Pipeline> retval =
+        pipelines.stream().filter(pipelinePredicate).collect(Collectors.toList());
+
+    log.debug("returning {} of {} total pipeline(s)", retval.size(), pipelines.size());
+
+    return sortPipelines(retval);
+  }
+
+  private List<Pipeline> sortPipelines(List<Pipeline> pipelines) {
     pipelines.sort(
         (p1, p2) -> {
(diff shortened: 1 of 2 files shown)


front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -202,11 +202,36 @@ public Collection<Pipeline> getTriggeredPipelines(
...
+    if (!enabledPipelines) {
+      return sortPipelines(pipelines);
+    }
+
+    Predicate<Pipeline> pipelinePredicate =
+        pipeline -> {
+          // pipeline.getDisabled may be null, so check that before comparing.  If
+          // pipeline.getDisabled is null, the pipeline is enabled.
+          boolean pipelineEnabled =
+              (pipeline.getDisabled() == null) || (pipeline.getDisabled() == false);
```
jasonmcintosh, line 221: NIT:  MIght use the Optional stuff here :)

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -202,11 +202,36 @@ public Collection<Pipeline> getTriggeredPipelines(
   public List<Pipeline> listByApplication(
       @PathVariable(value = "application") String application,
       @RequestParam(value = "pipelineNameFilter", required = false) String pipelineNameFilter,
-      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh) {
+      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh,
+      @RequestParam(required = false, value = "enabledPipelines", defaultValue = "false")
+          Boolean enabledPipelines) {
```
dbyron-sf, line 207: Pretty sure this can be lower-case-b boolean since there's a default.
```suggestion
          boolean enabledPipelines) {
```

front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -202,11 +202,45 @@ public Collection<Pipeline> getTriggeredPipelines(
   public List<Pipeline> listByApplication(
       @PathVariable(value = "application") String application,
       @RequestParam(value = "pipelineNameFilter", required = false) String pipelineNameFilter,
-      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh) {
+      @RequestParam(required = false, value = "refresh", defaultValue = "true") boolean refresh,
+      @RequestParam(required = false, value = "pipelineStateFilter", defaultValue = "all")
```
dbyron-sf, line 206: What are the other possible values for pipelineStateFilter?  Unless there's something super obvious coming down the road soon, I'd rather use a `boolean`.  Plus also, even though `refresh` does it, I don't think we need/want to specify value if it's the same as the argument name.

christosarvanitis: It is `all`,`enabled`,`disabled`. The reason i switched from boolean is that i can see a useful predefined filter in Deck -> Gate for getting only the enabled or the disabled pipelines. 

But we can always leave this as a future improvement if it is preferred to keep it cleaner for now.

dbyron-sf: Even if this info comes from gate (which would mean a change to https://github.com/spinnaker/orca/pull/4819 right?), if those are the values, then a Boolean is enough, isn't it?

christosarvanitis: Gate directly calls Front50 for the pipelineConfigIds and then calls Orca to fetch the executions for those pipelineConfigIds.So both changes are needed.

Yes,sure i can change to Boolean with null fetching everything and true|false for the enabled|disabled

christosarvanitis: Pushed the Boolean change here and in the orca PR

jasonmcintosh (approved):

PR #1525, opened 2025-01-23
fix(migrations): Fix wiring secondarySqlStorage service bean

christosarvanitis:
Fix wiring of the secondarySqlStorageService to allow SQL to SQL migrations using the CompositeStorageService

The secondaryJooq is only created when different dialect exists between the default and secondary connection pool which is not the case for Front50 migration between two SQL persistent stores


radu-dragon (approved):

front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -58,11 +56,11 @@ class SqlConfiguration {
     )
 
   @Bean
-  @ConditionalOnBean(name = ["secondaryJooq"])
+  @ConditionalOnProperty("sql.enabled", "sql.secondary.enabled")
   fun secondarySqlStorageService(
     objectMapper: ObjectMapper,
     registry: Registry,
-    @Qualifier("secondaryJooq") jooq: DSLContext,
+    jooq: DSLContext,
```
jasonmcintosh, line 63: rmm, wont' this use the primary DSLContext without the qualifier?  NOT dug yet just... this seems to be an interesting change...

jasonmcintosh:
Wondering if we couldn't get a spring test to verify this functionality as well - to make sure we're able to load this!

front50-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
christosarvanitis: secondaryJooq is conditionally created when the dialect doesnt match the default pool. I have pushed a change that when secondaryJooq exists it is used as the effective DSLContext. When it doesnt exist then it defaults to the jooq DSLContext.

jasonmcintosh:
Need to get tests more stable, but I think this looks good.  Re-review once tests are adjusted to handle startup latency time.

christosarvanitis:
> Need to get tests more stable, but I think this looks good. Re-review once tests are adjusted to handle startup latency time.

@jasonmcintosh pushed some stability changes to the integration tests. Feel free to merge if you are ok with the changes

